PR #2, opened 2016-05-26
Formatting

august782:
Some change to formatting for a switch statement and some small spacing and capitalization changes.


alexgyori:
Could you check why the checkstyle doesn't prevent the switch error? It seems that the indentation modules is installed and it requires some case indentation but it's misconfigured.

Similarly, for the space before comma, that should also be caught by out checkstyle. Can you fix the checkstyle?

PR #3, opened 2016-05-26
Adding checkstyle plugin to the pom.xml for core

august782:
In case you actually want to check the style in nondex-core, where the tests are, adding the checkstyle plugin to the pom.xml for that module.

Note, this triggers many violations of the checkstyle you may want to fix.

### DIFF
--- a/nondex-core/pom.xml
+++ b/nondex-core/pom.xml
@@ -14,19 +14,50 @@
 
   <build>
     <plugins>
-        <plugin>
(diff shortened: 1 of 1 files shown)


alexgyori:
Will accept this once we remove the java.* source files since those likely trigger many of the violations.

alexgyori:
I updated things in one atomic commit to fix both the errors and add also checkstyle.

PR #13, opened 2016-06-02
Added a few things to the root pom.xml per maven central requirements

owolabileg:


alexgyori:
@owolabileg  @august782 @darko-marinov @azy2 make sure this is the info you want in the pom about you. This info is needed to make the plugin available in maven central.

august782:
I think the URL should point to the GitHub page, as opposed to front page of Mir.

owolabileg:
@august782, done

PR #17, opened 2016-06-03
Adding logic to return early when debugging found single choice point

august782:
Checks when a debugging results in the start and end being the same and returning the moment that happens, as to avoid searching other executions to get the same thing (requires making numChoices public).

Also requires doing some wrapping to make a new Configuration to represent that debugged failure that has the updated start and end values, otherwise those values stay the same as limits.

### DIFF
--- a/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/DebugTask.java
+++ b/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/DebugTask.java
@@ -114,10 +114,14 @@ public class DebugTask {
             if (this.failsOnDry(config)) {
                 Configuration failingConfig = this.startDebugBinary(config);
 
+                // If debugged down to single choice point, then go ahead and return that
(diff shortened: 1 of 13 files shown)


nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/DebugTask.java
```diff
@@ -114,10 +114,14 @@ private Configuration debugWithConfigurations(Set<Configuration> failingConfigur
...
+                // If debugged down to single choice point, then go ahead and return that
+                if (failingConfig.numChoices() == 0) {
+                    return failingConfig;
+                }
+                // Otherwise should go on until finding better one
                 if (debConfig == null) {
                     debConfig = failingConfig;
-                }
-                if (failingConfig.hasLessChoicePoints(debConfig)) {
+                } else if (failingConfig.hasLessChoicePoints(debConfig)) {
```
alexgyori: it should be fewer not less :).

PR #30, opened 2016-06-08
fix multi-logging

alexgyori:


alexgyori:
@owolabileg @august782  This fixes the bug where I was logging multiple times a failing test.

august782:
I see that this commit stops adding the listener after the first execution in the same JVM, but do you happen to know why it's being invoked multiple times?

alexgyori:
Because we rerun surefire multiple times. And I was passing to every new execution the listener, and it seems every time I pass it it registers a new listener object. So if I only pass it the first time, the listener will be registered once and every failing test will print once.

Sent from my smartphone.

On Jun 8, 2016, at 12:04 PM, august782 <notifications@github.com<mailto:notifications@github.com>> wrote:


I see that this commit stops adding the listener after the first execution in the same JVM, but do you happen to know why it's being invoked multiple times?

owolabileg:
So why do we see the logging multiple times in some projects but not in others?

alexgyori:
@owolabileg That is a good question. :)
I am not sure if/when/where we do. We would have to recheck/rerun. 
I was not sure last night what to test on, i.e., which projects don't print duplicates since I thought we did not double check that those old results do not print duplicates..

august782:
So I think by itself, the changes look reasonable to me, in that it prevents adding extra listeners during the same JVM execution with multiple executions of Surefire with different seeds (I'm assuming you tried printing out the listener value during a run to see that before it was just appending things?).

Concerning when a listener does not output duplicates, I think the project that did not exhibit this was commons-cli. One possibility is that every time only one seed causes a failure, so no duplicates can be printed out, but this seems extremely unlikely.

PR #31, opened 2016-06-08
fixed NPE when some file is not in the jar

alexgyori:
This actually happened when running on an old jar without the .md5 files in it. It could happen also when a jar does not have all the class files in it.


alexgyori:
@azy2 Please check this is correct. I'll merge since I need the fix but feel free to improve it.

PR #39, opened 2016-06-10
Debugging and bugfixing multilogging

alexgyori:


alexgyori:
Outdated. Was fixed already.

PR #55, opened 2016-06-15
Improved the look of test_results.html

azy2:


alexgyori:
@azy2 my browser was displaying weird characters instead of the checkmark, so I changed the code  a bit.

azy2:
Can you also take a look at line 161. I believe that is printing the same seed for every run but I don't know where the individual seeds for each run can be found.

alexgyori:
@azy2 You are iterating over executions. Each execution has a configuration. You should print the one you are iterating over.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
```diff
@@ -167,7 +167,7 @@ private void generateHtml(Set<String> allFailures, Configuration config) {
         html += "<th>Test Name</th>";
         for (int iter = 0; iter < this.executions.size(); iter++) {
             html += "<th>";
-            html += "" + config.seed;
+            html += "" + this.executions.get(iter).getConfiguration().seed;
```
alexgyori: you should use an enhanced foreach instead of this indexing?

alexgyori: Actually, keep it as is, but add above the seed what was already there "Run i" maybe?

PR #64, opened 2016-06-21
Hash maps

azy2:
Don't merge yet, I still haven't added IdentityHashMap or WeakHashMap


nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/HashMapShufflingAdder.java
```diff
@@ -211,31 +159,30 @@ public void visitInsn(int opcode) {
...
-
-            return original;
+        if (name.equals("next" + type)) {
+            nextTypeProp.access = access;
+            nextTypeProp.name = name;
+            nextTypeProp.desc = desc;
+            nextTypeProp.signature = signature;
+            nextTypeProp.exceptions = exceptions;
+
+            if (baseName.equals("java/util/WeakHashMap")) {
```
alexgyori: @azy2 This is worth fixing. (be robust to NPE) I am going to disable the other string literal checks/warnings.

PR #80, opened 2016-09-04
Update appveyor.yml

alexgyori:
Run 'mvn clean verify' to run integration tests.


alexgyori:
@DrRacket, the change that fixed the integration tests is the one that changes ":" to File.pathSeparator.
We need that because on Windows for example, you separate paths in the cp by ';'.

DrRacket:
@alexgyori I see. In #109 , I've made the similar change.

PR #84, opened 2016-10-06
refactoring the NonDex code that controls exploration

### LINKED ISSUE #83: Refactor ControlNondeterminism.java
There are too many ifs and the logic is hard to understand.
We just need a common API to create a nondeterministic choice and enable the choice or disable it; currently it's all over the place with error-prone if checks.

alexgyori:
Fixes #83

### DIFF
--- a/nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
+++ b/nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
@@ -105,7 +105,7 @@ public class NonDex {
         if (this.shouldExplore()) {
             return newOrder;
         } else {
-            return new ArrayList<T>(originalOrder);
+            return originalOrder;
         }
     }
 
--- a/nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
+++ b/nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
@@ -29,6 +29,7 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 package edu.illinois.nondex.common;
 
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Random;
@@ -132,7 +133,9 @@ public class NonDex {
     }
 
     private String getInvocationElement() {
-        for (StackTraceElement el : Thread.currentThread().getStackTrace()) {
+        // We need to skip here the first element of the stack since that is in the Thread class 
+        StackTraceElement[] stack = Thread.currentThread().getStackTrace();
+        for (StackTraceElement el : Arrays.asList(stack).subList(1, stack.length - 1)) {
             if (!el.getClassName().startsWith("edu.illinois.nondex")) {
                 Logger.getGlobal().log(Level.FINE, "The invocation element is: " + el.toString());
                 return el.toString();
--- a/nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
+++ b/nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
@@ -38,8 +38,6 @@ import java.util.List;
 import java.util.Random;
 import java.util.logging.Level;
 
-import edu.illinois.nondex.shuffling.ControlNondeterminism;
-
 public class NonDex {
 
     private int opportunityCount;
--- a/nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
+++ b/nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
@@ -28,12 +28,17 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 
 package edu.illinois.nondex.common;
 
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.StandardOpenOption;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Random;
 import java.util.logging.Level;
+
+import edu.illinois.nondex.shuffling.ControlNondeterminism;
 
 public class NonDex {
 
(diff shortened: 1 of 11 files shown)


alexgyori:
@august782 @azy2 @darko-marinov  I need a review to be able to merge :).

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
```diff
@@ -189,40 +120,34 @@ private static void printStackTraceIfUniqueDebugPoint() {
...
                         StandardOpenOption.CREATE, StandardOpenOption.APPEND);
-                Files.write(ControlNondeterminism.config.getInvocationsPath(),
+                Files.write(nondex.getConfig().getInvocationsPath(),
                         ("SHUFFLES:" + localShufflesCount + String.format("%n")).getBytes(),
                         StandardOpenOption.APPEND);
             } catch (IOException ioe) {
-                Logger.getGlobal().log(Level.SEVERE, "Exception when printing shuflling counts in shutdown hook.", ioe);
+                Logger.getGlobal().log(Level.SEVERE,
+                        "IOException when printing shuflling counts in shutdown hook.", ioe);
+            } catch (Throwable ex) {
```
august782: Spelling mistake, shuflling -> shuffling

nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
```diff
@@ -0,0 +1,163 @@
...
+        if (this.shouldExplore()) {
+            return result;
+        }
+
+        return identityElement;
+    }
+
+    public int getInteger() {
+
+        return this.getInteger(0);
```
august782: Always returns 0 when not exploring? What's the intended use case?

nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    public <T> List<T> permutation(List<T> originalOrder) {
+
+        if (originalOrder == null) {
+            throw new IllegalArgumentException("originalOrder is null");
+        }
+
+        List<T> newOrder = new ArrayList<T>(originalOrder);
+        Collections.shuffle(newOrder, this.getRandom());
```
august782: Why not move this shuffle into the if block so there is no need to make a new ArrayList again in the else block?

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
```diff
@@ -93,93 +70,47 @@ private static Random getRandomnessSource(String source) {
...
 
         List<T> newOrder = Arrays.asList(originalOrder);
-        newOrder = ControlNondeterminism.internalShuffle(newOrder, ControlNondeterminism.getSource());
+
+        newOrder = nondex.permutation(newOrder);
+
         newOrder.toArray(originalOrder);
 
         // return in place
         return originalOrder;
```
august782: With List it returns new List, while with array it returns the original modified array. Is there a reason they should be different in this regard?

nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
```diff
@@ -0,0 +1,163 @@
...
+    private boolean shouldExploreForInstance() {
+        return this.opportunityCount >= this.config.start
+                && this.opportunityCount <= this.config.end;
+    }
+
+    public Configuration getConfig() {
+        return this.config;
+    }
+
+    public int getpossibleExplorations() {
```
august782: Method name should be camel-cased.

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
alexgyori: Well... that method is not part of the NonDex interface. There are two reasons why we don't create a new array:
1) It would require using reflection to find the type of the array I need to create (originalOrder.getClass() and then create some object of that type again via reflection...).
2) it's safe to do so for what we are doing, since all APIs are saying that the user can modify the return value.

nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
alexgyori: Fixed.

nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
alexgyori: But, I actually don't need to create a new array in the else, I can just return originalOrder, right?

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
```diff
@@ -93,136 +68,53 @@ private static Random getRandomnessSource(String source) {
...
                         StandardOpenOption.CREATE, StandardOpenOption.APPEND);
-                Files.write(ControlNondeterminism.config.getInvocationsPath(),
+                Files.write(nondex.getConfig().getInvocationsPath(),
                         ("SHUFFLES:" + localShufflesCount + String.format("%n")).getBytes(),
                         StandardOpenOption.APPEND);
             } catch (IOException ioe) {
-                Logger.getGlobal().log(Level.SEVERE, "Exception when printing shuflling counts in shutdown hook.", ioe);
+                Logger.getGlobal().log(Level.SEVERE,
+                        "IOException when printing shuffling counts in shutdown hook.", ioe);
+            } catch (Throwable ex) {
```
august782: Codacy keeps flagging this (and sending emails...), is there some Throwable that can be thrown here that Exception would not catch?

alexgyori: I can disable this warning, presumably if I push again it should not warn us again :)...

PR #92, opened 2016-10-15
fix bug of hashmap iterator remove

DrRacket:
Add the one line change in the HashIteratorShuffler.java.
Modified the HashIteratorShufflerASMDump.java

I looked at the source code of WeakHashMap and it turns out that they used different names for the variables. So I added a if-else in the ASMDump file. And I modified the HashMapShufflingAdder since that variable in the WeakHashMap is private while the "current" in the HashMap is public...

Added a simple test case for both HashMap and WeakHashMap.
Also fixed a "bug" in the test case of the WeakHashMap.

### DIFF
--- a/nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/HashIteratorShufflerASMDump.java
+++ b/nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/HashIteratorShufflerASMDump.java
@@ -70,6 +70,14 @@ public class HashIteratorShufflerASMDump implements Opcodes {
             mv.visitMethodInsn(INVOKESTATIC, "edu/illinois/nondex/shuffling/ControlNondeterminism", "shuffle", "(Ljava/util/List;)Ljava/util/List;", false);
             mv.visitVarInsn(ASTORE, 3);
             mv.visitVarInsn(ALOAD, 0);
+            mv.visitFieldInsn(GETFIELD, baseName + "$HashIterator$HashIteratorShuffler", "hashIter", "L" + baseName + "$HashIterator;");
+            mv.visitInsn(ACONST_NULL);
+            if(baseName.equals("java/util/WeakHashMap")){
+                mv.visitFieldInsn(PUTFIELD, baseName + "$HashIterator", "lastReturned", "L" + baseName + "$" + type + ";");
+            } else {
+                mv.visitFieldInsn(PUTFIELD, baseName + "$HashIterator", "current", "L" + baseName + "$" + type + ";");
+            }
+            mv.visitVarInsn(ALOAD, 0);
             mv.visitVarInsn(ALOAD, 3);
             mv.visitMethodInsn(INVOKEINTERFACE, "java/util/List", "iterator", "()Ljava/util/Iterator;", true);
             mv.visitFieldInsn(PUTFIELD, baseName + "$HashIterator$HashIteratorShuffler", "iter", "Ljava/util/Iterator;");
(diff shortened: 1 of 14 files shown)


nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/HashMapShufflingAdder.java
```diff
@@ -123,7 +123,8 @@ public void visitEnd() {
 
     @Override
     public FieldVisitor visitField(int access, String name, String desc, String signature, Object value) {
-        if ("java/util/WeakHashMap".equals(baseName) && ("expectedModCount".equals(name) || "entry".equals(name))) {
+        if ("java/util/WeakHashMap".equals(baseName)
+                && ("expectedModCount".equals(name) || "entry".equals(name) || "lastReturned".equals(name))) {
```
alexgyori: Why? Please add a comment. What is this lastReturned?

nondex-test/src/test/java/edu/illinois/nondex/core/WeakHashMapTest.java
```diff
@@ -48,6 +50,17 @@ public void weakHashMapTest() {
             whm.put(i, i);
         }
 
-        assertThat("{0=0, 2=2, 3=3, 4=4, 5=5, 6=6, 7=7, 8=8, 9=9}", not(equalTo(whm.toString())));
```
alexgyori: this was a bad test... good fix.

nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/HashIteratorShufflerASMDump.java
```diff
@@ -70,6 +70,14 @@
             mv.visitMethodInsn(INVOKESTATIC, "edu/illinois/nondex/shuffling/ControlNondeterminism", "shuffle", "(Ljava/util/List;)Ljava/util/List;", false);
             mv.visitVarInsn(ASTORE, 3);
             mv.visitVarInsn(ALOAD, 0);
+            mv.visitFieldInsn(GETFIELD, baseName + "$HashIterator$HashIteratorShuffler", "hashIter", "L" + baseName + "$HashIterator;");
+            mv.visitInsn(ACONST_NULL);
+            if("java/util/WeakHashMap".equals(baseName)){
+                mv.visitFieldInsn(PUTFIELD, baseName + "$HashIterator", "lastReturned", "L" + baseName + "$" + type + ";");
```
alexgyori: Was this another bug? I don't remember discussing this bug with you. Please edit your pull request describing what this is supposed to address. It's fine if you submit PRs with little to no text if I know what they're about... but I don't think we discussed this. I'm guessing what this is, but please add some description for posterity.

alexgyori: The WeakHashMap source is also missing.

alexgyori: You do not need this if. Look in Instrumenter.java on line 140 it invokes the dump(...) method with certain arguments. There, the variable name is passed in, i.e., current or entry. I believe you need to change that call, in the branch corresponding to WeakHashMap to call with "lastReturned". That way you'll not have this if here but rather just pass from outside the right argument.

alexgyori: Also, please add the source for WeakHashMap. It seems that the source for IdentityHashMap is also missing... please check what else is missing.

DrRacket: The source code of WeakHashMap, IdentityHashMap, PriorityQueue are missing. I will add them soon.

alexgyori: I see you added the sources. But they are not the modified sources to refer to the shuffler. 
1) Can you add a comment where the shuffler would be used?
2) As I said earlier, remove this IF and pass the name of the field as argument in the Instrumenter, i.e., lastReturned.

PR #95, opened 2016-10-31
fix weakhashmap

DrRacket:
Rewrite the logic for weakhashmap. Fixed the remove bug.
now the weakhashmap shuffling logic does not use hashiteratorshuffler.


nondex-core/src/main/java/java/util/WeakHashMap.java
```diff
@@ -763,6 +763,8 @@ public String toString() {
         private Entry<K,V> lastReturned = null;
         private int expectedModCount = modCount;
 
+//        HashIteratorShuffler shuffler;
```
alexgyori: Why this comment?

nondex-core/src/main/java/java/util/WeakHashMap.java
```diff
@@ -775,13 +777,33 @@ public String toString() {
...
+        public Entry<K, V> nextEntry() {
+            if (modCount != expectedModCount)
+                throw new ConcurrentModificationException();
+            lastReturned = iter.next();
+            currentKey = lastReturned.get();
+            return lastReturned;
+        }
+
+        public boolean originalHasNext() {
+            Entry<K,V>[] t = table;
```
alexgyori: You should not need to edit the originalHasNext. Why do you need this?

nondex-core/src/main/java/java/util/WeakHashMap.java
DrRacket: Will delete this.

nondex-core/src/main/java/java/util/WeakHashMap.java
DrRacket: I actually didn't modify it. You can look at line 783 and line 806 and you will find that this is because of how the "diff" works.

nondex-test/src/test/java/edu/illinois/nondex/core/IdentityHashMapTest.java
```diff
@@ -1,53 +0,0 @@
-/*
```
alexgyori: why did you remove this here? Seems unrelated to the PR.

nondex-core/src/main/java/java/util/WeakHashMap.java
```diff
@@ -763,6 +763,7 @@ public String toString() {
         private Entry<K,V> lastReturned = null;
         private int expectedModCount = modCount;
 
+
```
DrRacket: Will delete.

nondex-test/src/test/java/edu/illinois/nondex/core/IdentityHashMapTest.java
DrRacket: Yeah.. Should be in another PR.. I will discard the change.

PR #96, opened 2016-11-06
Instrumenter tests

alexgyori:
The tests in nondex-instrumentation were under src/tests/edu/... instead of src/tests/java/edu/...
Due to this, they were not getting compiled, nor run; 

There are fixes to move them to the right directory, make them compile, and also fix a bug in the instrumenter code that would throw a NPE when a class is not present in the jar that is given as input (typically rt.jar). If this test was in the right place, the bug would've been found earlier.


nondex-instrumentation/src/test/java/edu/illinois/nondex/instr/InstrumenterTest.java
```diff
@@ -28,33 +28,39 @@ a copy of this software and associated documentation files (the
...
-            Instrumenter.instrument(Paths.get("resources", "empty.jar"), Paths.get("resources", "emptyOut.jar"));
-            fail("Expected an IOException to be thrown");
-        } catch (IOException exc) {
-            assertThat(exc.getMessage(), is("zip file is empty"));
-        }
+    public void emptyZipTest() throws NoSuchAlgorithmException, IOException {
+        Path outPath = Paths.get("resources", "emptyOut.jar");
+        Instrumenter.instrument(Paths.get("resources", "empty.jar").toString(),
+                                outPath.toString());
+        assertTrue(Files.exists(outPath, LinkOption.NOFOLLOW_LINKS));
```
DrRacket: After running this test, will the emptyOut.jar be created? Should we remove it after the run?

alexgyori: I'll change it so that the test cleans up after itself.

PR #99, opened 2016-11-12
Fix IdentityHashMap remove bug

DrRacket:
Didn't modify the remove function.
When initializing the iterator, both the list of keys and the list of key indices are created, and the list of key indices will be shuffled.
And when calling nextIndex, the function will loop through the table which holds everything and find the current index of the key which should be returned.
Also removed the IdentityHashMapTest since it's a bad test case.


.travis.yml
```diff
@@ -7,3 +7,6 @@ os:
 before_install:
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; brew tap caskroom/cask; brew install Caskroom/cask/java; fi
     - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then jdk_switcher use oraclejdk8; fi
+
```
alexgyori: remove this. We only added it for debugging.

PR #100, opened 2016-11-18
Fix priorityQueue bugs, add parameterized tests

DrRacket:
The old parameterized tests are removed.


nondex-core/src/main/java/java/util/PriorityQueue.java
```diff
@@ -524,13 +524,40 @@ public boolean contains(Object o) {
          */
         private int expectedModCount = modCount;
 
+        private List<E> elements;
+        private Iterator<E> iter;
+        
+        {
```
alexgyori: Put this in a constructor; let's not use the non-static initializer.

nondex-core/src/main/java/java/util/PriorityQueue.java
```diff
@@ -524,13 +524,40 @@ public boolean contains(Object o) {
...
+        }
+
+        public boolean hasNextOrig() {
             return cursor < size ||
                 (forgetMeNot != null && !forgetMeNot.isEmpty());
         }
 
         @SuppressWarnings("unchecked")
-        public E next() {
+        public E nextOrig() {
```
alexgyori: call it originalNext and originalHasNext below; let's follow the same convention everywhere.

nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/PriorityQueueShufflingAdder.java
```diff
@@ -176,35 +172,18 @@ public void visitEnd() {
...
-            hasNextProp.desc = desc;
-            hasNextProp.signature = signature;
-            hasNextProp.exceptions = exceptions;
-
-            MethodVisitor original = super.visitMethod(access, "hasNextOrig", desc, signature, exceptions);
-
-            return original;
+            return super.visitMethod(access, "hasNextOrig", desc, signature, exceptions);
         }
         if ("next".equals(name)) {
```
alexgyori: Add an if here to check if this class has already a constructor, i.e., <init> method. If so, we'd need to change the current instrumentation, which just generates an <init> method from scratch, to add the code to the existing <init> methods.

nondex-test/src/test/java/edu/illinois/nondex/core/QueueTest.java
```diff
@@ -0,0 +1,136 @@
...
+    public void setUp() {
+        queue.clear();
+        for (int ind = 0; ind < 10; ind++) {
+            queue.add(ind);
+        }
+    }
+
+    @Test
+    public void testShuffling() {
+        if (queue instanceof PriorityBlockingQueue<?>) {
```
alexgyori: Put specifically at TODO comment here so that we know this if is not meant to be permanent.

nondex-test/src/test/java/edu/illinois/nondex/core/QueueTest.java
```diff
@@ -0,0 +1,136 @@
...
+
+    @Test(expected = IllegalStateException.class)
+    public void testRemoveWhenEmpty() {
+        queue.clear();
+        Iterator<Integer> iter = queue.iterator();
+        iter.remove();
+    }
+
+    @Test(expected = IllegalStateException.class)
+    public void testRemoveWhenNotEmpty() {
```
alexgyori: change testRemoveWhenNotEmpty -> testRemoveWithoutCallingNextBeforeIt (or something like that, to illustrate what the test is supposed to test :)

alexgyori:
Check the log, for example: https://api.travis-ci.org/jobs/177821426/log.txt?deansi=true
It contains your warning that init already exists. Why does it contain it? :)
Also prints twice, which indicates that there are two constructors already in the bytecode.
I would just change the implementation to just add the code you have to all constructors, rather than generating one from scratch and overwriting existing ones.

nondex-test/src/test/java/edu/illinois/nondex/core/QueueTest.java
```diff
@@ -86,7 +87,7 @@ public void testRemoveWhenEmpty() {
     }
 
     @Test(expected = IllegalStateException.class)
-    public void testRemoveWhenNotEmpty() {
+    public void testRemoveProirToNext() {
```
alexgyori: there is no next in this test... prior means before and it sort of implies there will be a call to next.
Just say: testRemoveWithoutCallingNextBefore
It's ok for the method name to be long, it's not as if anybody would call this method ever and needs to type its name.

alexgyori:
Looks good to merge to me. @DrRacket is this done?

DrRacket:
Yes.

DrRacket:
@alexgyori Yes. Please merge it.

PR #104, opened 2016-11-28
Fix toString bug in PriorityBlockingQueue

DrRacket:
The toString method is not shuffled by NonDex before.
Since the PriorityBlockingQueue toString method didn’t use the iterator
of PriorityBlockingQueue, this method needs to be instrumented.
PriorityBlockingQueue is originally classified as a standard class to
instrument because only toArray method is instrumented. I moved the
class to special class to instrument and instrumented both toString and
toArray methods.

### DIFF
--- a/nondex-core/src/main/java/java/util/concurrent/PriorityBlockingQueue.java
+++ b/nondex-core/src/main/java/java/util/concurrent/PriorityBlockingQueue.java
@@ -731,7 +731,7 @@ public class PriorityBlockingQueue<E> extends AbstractQueue<E>
         }
     }
 
-    public String toString() {
+    public String originalToString() {
         final ReentrantLock lock = this.lock;
         lock.lock();
         try {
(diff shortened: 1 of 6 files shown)


nondex-core/src/main/java/java/util/concurrent/PriorityBlockingQueue.java
```diff
@@ -752,6 +752,29 @@ public String toString() {
         }
     }
 
+    public String toString() {
```
alexgyori: Is there a legitimate reason why you can't call the iterator from here, rather than calling NonDex directly?
I would just change the for loop to for (Object e : iterator()) - or something like that.

nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/ClassVisitorShufflingAdder.java
```diff
@@ -66,7 +66,6 @@ a copy of this software and associated documentation files (the
 
         apisReturningShufflableArrays.add("java/util/PriorityQueue.toArray");
 
-        apisReturningShufflableArrays.add("java/util/concurrent/PriorityBlockingQueue.toArray");
```
alexgyori: Are you saying toArray uses the iterator() method?

nondex-core/src/main/java/java/util/concurrent/PriorityBlockingQueue.java
DrRacket: I just try to minimize the changes I made so it will be possibly safer so I don't introduce more bugs..
I agree that using an iterator should also work but I didn't test that and honestly I don't know why the author of this method didn't use the iterator.
Maybe you should ask the developers why they choose not to use the iterator here. :)

PR #105, opened 2016-11-29
Added Systematic Exploration feature

maronesy:
Hi Alex,

I have added the systematic exploration feature to NonDex.  I have added the following new classes:

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
nondex-common/src/main/java/edu/illinois/nondex/common/StackElement.java

And modfied the following classes:

nondex-common/src/main/java/edu/illinois/nondex/common/Mode.java
nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java

I kindly ask that you merge my changes to NonDex.  Please let me know if there is anything that needs to be modified/corrected before accepting my PR.

Thanks,
Souhayl


nondex-test/src/test/java/edu/illinois/nondex/common/SystematicRandomTest.java
```diff
@@ -26,28 +26,29 @@ a copy of this software and associated documentation files (the
...
-import static org.hamcrest.core.IsEqual.equalTo;
-import static org.hamcrest.core.IsNot.not;
-import static org.junit.Assert.assertThat;
+package edu.illinois.nondex.common;
 
+import java.lang.reflect.Array;
 import java.util.ArrayList;
-import java.util.IdentityHashMap;
+import java.util.Collections;
+import java.util.List;
```
alexgyori: Please remove all unused imports.

nondex-test/src/test/java/edu/illinois/nondex/common/SystematicRandomTest.java
```diff
@@ -26,28 +26,29 @@ a copy of this software and associated documentation files (the
...
-
-import static org.hamcrest.core.IsEqual.equalTo;
-import static org.hamcrest.core.IsNot.not;
-import static org.junit.Assert.assertThat;
+package edu.illinois.nondex.common;
 
+import java.lang.reflect.Array;
 import java.util.ArrayList;
-import java.util.IdentityHashMap;
+import java.util.Collections;
```
alexgyori: Remove all unused imports.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+                    bufferedWriter.write(lastAndMax);
+                    bufferedWriter.newLine();
+                }
+                bufferedWriter.close();
+                return;
+            }
+            for (StackElement ch: choice) {
+                ex.add(ch.getExplore());
+            }
+            if (!ex.contains(true)) {
```
alexgyori: You seem to be using this ex only as a flag - if it was ever set, you go into this if. Why not just use a boolean flag, for example on line 134 you could do: flag = flag || ch.getExplore();

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
```diff
@@ -56,7 +56,7 @@ public static Configuration getConfiguration() {
     }
 
     public static <T> List<T> shuffle(List<T> originalOrder) {
-        return nondex.getPermutation(originalOrder);
+        return nondex == null ? originalOrder : nondex.getPermutation(originalOrder);
```
alexgyori: Why the check? This is never null.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
```diff
@@ -108,6 +103,20 @@ public void execute() throws MojoExecutionException, MojoFailureException {
 
     }
 
+    private void executeHelper(int seedNum) {
+        NonDexSurefireExecution execution =
```
alexgyori: Here you would need to make some changes, so that you pass nondexStart and nondexEnd.

nondex-test/src/test/java/edu/illinois/nondex/common/SystematicRandomTest.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.junit.Test;
+
+
+public class SystematicRandomTest {
+
+//    @Test
```
alexgyori: You should not commit such tests.

alexgyori:
The most important is to fix the build. You can look into the Travis and AppVeyor logs to see what is failing.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+                        int last = Integer.parseInt(choiceValues[0].toString());
+                        int max = Integer.parseInt(choiceValues[1].toString());
+                        boolean explore = Boolean.parseBoolean(choiceValues[2].toString());
+                        StackElement lastMax = new StackElement( last, max, explore );
+                        choice.push(lastMax);
+                    }
+                }
+            }
+        } else {
+            System.err.println("The env variable does not exist. File could not be created");
```
DrRacket: @alexgyori Should we avoid using System.err.println?
Checkstyle is only checking for System.out.println now.

alexgyori: We should. You never used it though so I didn't comment.
All printing should be done through logging and no other way - that enables us to filter and selectively turn off and on printing, and easily redirect it to some file if we wanted.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+            System.exit(1);
```
august782: This appears to be the problem that's causing the build to fail on Travis (and any machine where the environment variable is not defined). The System.exit(1) is invoked since the environment variable does not exist, and due to how the tests are doing things with threads and stuff, it just hangs, leading to the timeout on Travis. There really should be no System.exit(1) in the code, some exception should be thrown and propagated upwards.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.Stack;
+
+public class SystematicRandom extends Random {
+    public static int STARTING_COUNT = 59;
+    int replayIndex;
+    int count;
+    Stack<StackElement> choice;
```
DrRacket: Use ArrayDeque instead of Stack.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+            StackElement choiceNums = new StackElement(last, max, explore);
+            choice.push(choiceNums);
+        }
+        replayIndex++;
+        return last;
+    }
+
+    public void endRun() throws IOException {
+        while (!choice.isEmpty()) {
+            ArrayList<Object> ex = new ArrayList<>();
```
DrRacket: Why do you need this list? Why ArrayList<Object>?

nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
```diff
@@ -68,6 +70,10 @@ public static NonDex getInstance() {
         return globalInstance;
     }
 
+    public SystematicRandom getSysRandom() {
```
maronesy: The getRandom method is private and part of the edu.illinois.nondex.common package, since I need to get systematicRandom in package edu.illinois.nondex.shuffling I added a new public method purely for getting systematicRandom.  If its ok for me to change the getRandom method to public I will do so and remove the getSysRandom method.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Random;
+import java.util.Stack;
+
+public class SystematicRandom extends Random {
+    public static int STARTING_COUNT = 59;
+    int replayIndex;
+    int count;
```
maronesy: I removed count.

nondex-common/src/main/java/edu/illinois/nondex/common/NonDex.java
```diff
@@ -106,6 +112,7 @@ public int getInteger(int identityElement) {
 
         List<T> newOrder = new ArrayList<T>(originalOrder);
         Collections.shuffle(newOrder, this.getRandom());
+
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+    Stack<StackElement> choice;
+    String logFileName;
+    Charset utf8 = StandardCharsets.UTF_8;
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+        return last;
+    }
+
+    public void endRun() throws IOException {
+        while (!choice.isEmpty()) {
+            ArrayList<Object> ex = new ArrayList<>();
+            StackElement lastMax = choice.pop();
+            int last = lastMax.getLast();
+            int max = lastMax.getMax();
+            boolean explore = false;
```
maronesy: the explore here is used to flag whether the new element (last/max combination) we push to choice should be explored.  If you look inside the if branch you will see that we are changing last as well by incrementing it by one.  So I don't necessarily want the explore stored in the stack for the popped element but instead I want explore to be false unless we are pushing an element that we want to explore.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+            if (last < max - 1) {
+                last++;
+                if (choice.size() > STARTING_COUNT) {
+                    explore = true;
```
maronesy: This check is for the new element that we are pushing. Not for the ones that are already stored in the stack.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/StackElement.java
```diff
@@ -0,0 +1,54 @@
...
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE
+LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+package edu.illinois.nondex.common;
+
+public class StackElement {
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/StackElement.java
```diff
@@ -0,0 +1,54 @@
...
+
+    int last;
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/StackElement.java
```diff
@@ -0,0 +1,54 @@
...
+    int max;
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/StackElement.java
```diff
@@ -0,0 +1,54 @@
...
+    boolean explore;
+
+    public StackElement(int last, int max, boolean explore) {
+        this.last = last;
+        this.max = max;
+        this.explore = explore;
+    }
+
+    public int getLast() {
```
maronesy: Done.

nondex-test/src/test/java/edu/illinois/nondex/common/SystematicRandomTest.java
maronesy: It was committed by mistake! I apologize.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: ArrayDeque doest allow me to get specific elements based on Index which is something I am using.  I know I can use the iterator but stack already gives me a method get(index).

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: I am using ArrayList because in need something that does not have a fixed size something that is has a dynamic length.  The number of elements being added to ex may vary in each run.

nondex-common/src/main/java/edu/illinois/nondex/common/ConfigurationDefaults.java
```diff
@@ -60,6 +60,9 @@ a copy of this software and associated documentation files (the
     public static final String DEFAULT_END_STR = "9223372036854775807";
     public static final long DEFAULT_END = new Long(ConfigurationDefaults.DEFAULT_END_STR);
 
+    public static final String PROPERTY_LOG = "logFile";
+    public static final String DEFAULT_LOG_STR = "log";
```
alexgyori: log -> systematic.log
otherwise you may have duplicates and overwrite existing files - log is a fairly generic name that one could imagine is being used for many other things and there may be a file already called log created by users.

nondex-common/src/main/java/edu/illinois/nondex/common/ConfigurationDefaults.java
```diff
@@ -60,6 +60,9 @@ a copy of this software and associated documentation files (the
     public static final String DEFAULT_END_STR = "9223372036854775807";
     public static final long DEFAULT_END = new Long(ConfigurationDefaults.DEFAULT_END_STR);
 
+    public static final String PROPERTY_LOG = "logFile";
```
alexgyori: All nondex properties start with nondex*.
you should name the property nondexSystematicLogFile.
There may be other plugins that use logFile as an argument... then you won't be able to use nondex at the same time with that plugin.

nondex-common/src/main/java/edu/illinois/nondex/common/Configuration.java
```diff
@@ -55,6 +55,7 @@ a copy of this software and associated documentation files (the
 
     public final long start;
     public final long end;
+    public final String log;
```
alexgyori: What log is this?
log -> systematicLog
The names of your variables should be descriptive. Imagine that there may be other logs that the tool creates, e.g., a debug log.

nondex-common/src/main/java/edu/illinois/nondex/common/ConfigurationDefaults.java
alexgyori: PROPERTY_LOG -> PROPERTY_SYSTEMATIC_LOG

nondex-common/src/main/java/edu/illinois/nondex/common/ConfigurationDefaults.java
alexgyori: DEFAULT_LOG_STR -> DEFAULT_SYSTEMATIC_LOG

nondex-common/src/main/java/edu/illinois/nondex/common/ExplorationEntry.java
```diff
@@ -0,0 +1,54 @@
...
+
+package edu.illinois.nondex.common;
+
+public class ExplorationEntry {
+
+    private int current;
+    private int maximum;
+    private boolean shouldExplore;
+
+    public ExplorationEntry(int last, int max, boolean explore) {
```
alexgyori: Please rename last to current everywhere. Also, does your stack hold the current exploration, or the next exploration, i.e., the integers returned by systematic random in this run, or the ones that will be returned in the next run?

alexgyori: Please make sure this is changed for max and explore also, not only here, but in some other places around the code, in other classes, e.g., in SystematicRandom

nondex-test/src/test/java/edu/illinois/nondex/common/SystematicRandomTest.java
alexgyori: Okay, then either fill it in with some test, or remove it from this PR.

nondex-common/src/main/java/edu/illinois/nondex/common/Configuration.java
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/ExplorationEntry.java
maronesy: The stack holds the current exploration but the shouldExplore is used for the next exploration.

nondex-test/src/test/java/edu/illinois/nondex/common/SystematicRandomTest.java
maronesy: I have removed this test class for now.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+    public static int STARTING_COUNT = 59;
+    int replayIndex;
+    int count;
+    Stack<StackElement> choice;
+    String logFileName;
+    Charset utf8 = StandardCharsets.UTF_8;
+    List<String> lines;
+
+    public SystematicRandom() {
+        logFileName = System.getenv("logFileName");
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+        if (logFileName != null) {
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+            if (!file.exists()) {
+                choice = new Stack<StackElement>();
+                replayIndex = 0;
+                count = 0;
+            } else {
+                choice = new Stack<StackElement>();
+                try {
+                    lines = Files.readAllLines(Paths.get(logFileName));
+                } catch (IOException ioe) {
+                    ioe.printStackTrace();
```
maronesy: No, the code depends on systematic.log being read and than used.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+                }
+                Object[] choiceValues = new Object[3];
```
maronesy: Correct.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+                for (String element: lines) {
+                    String delimit = "[ ]+";
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+                    if (!element.isEmpty()) {
+                        choiceValues = element.split(delimit);
+                        int last = Integer.parseInt(choiceValues[0].toString());
+                        int max = Integer.parseInt(choiceValues[1].toString());
+                        boolean explore = Boolean.parseBoolean(choiceValues[2].toString());
+                        StackElement lastMax = new StackElement( last, max, explore );
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: I have removed the if and else branch by doing what you suggested.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: ex which has been renamed to exploreStorage is checking whether after the last pop we still have any elements that are set to true and should be explored.  I am storing all the shouldExplores from the stack in it and checking whether we still have any trues.

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
maronesy: If we remove this check when running mvn nondex:nondex we get a null pointed exception

DrRacket:
@maronesy Could you take a look at the Travis CI build log?
https://travis-ci.org/TestingResearchIllinois/NonDex/jobs/180858908#L2420
When running the tests, we are getting IOExceptions.
This might indicate that there are some issues in your SystematicRandom.endRun() method.

alexgyori:
@DrRacket That's a very good catch! Can you change our exception handling in the ControlNondeterminism.java so that we re-throw the IOException (In a different PR)?

DrRacket:
@alexgyori  Sure.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+    Charset utf8 = StandardCharsets.UTF_8;
+    List<String> lines;
+
+    public SystematicRandom() {
+        logFileName = System.getenv("logFileName");
+        if (logFileName != null) {
+            File file = new File(this.logFileName);
+            if (!file.exists()) {
+                choice = new Stack<StackElement>();
+                replayIndex = 0;
```
maronesy: I have removed the if/else and the fields.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+                count = 0;
+            } else {
+                choice = new Stack<StackElement>();
```
maronesy: I have removed if/else branch.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+            boolean explore = false;
+            if (last < max - 1) {
+                last++;
+                if (choice.size() > STARTING_COUNT) {
+                    explore = true;
+                }
+                StackElement lm = new StackElement(last, max, explore);
+                choice.push(lm);
+                replayIndex = 0;
+                if (Files.exists(Paths.get(logFileName))) {
```
maronesy: removed.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+                try {
+                    lines = Files.readAllLines(Paths.get(logFileName));
+                } catch (IOException ioe) {
+                    ioe.printStackTrace();
+                }
+                Object[] choiceValues = new Object[3];
+                for (String element: lines) {
+                    String delimit = "[ ]+";
+                    if (!element.isEmpty()) {
+                        choiceValues = element.split(delimit);
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+                if (choice.size() > STARTING_COUNT) {
+                    explore = true;
+                }
+                StackElement lm = new StackElement(last, max, explore);
+                choice.push(lm);
+                replayIndex = 0;
+                if (Files.exists(Paths.get(logFileName))) {
+                    Files.delete(Paths.get(logFileName));
+                }
+                BufferedWriter bufferedWriter = Files.newBufferedWriter(Paths.get(logFileName));
```
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/StackElement.java
```diff
@@ -0,0 +1,54 @@
...
+WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+*/
+
+package edu.illinois.nondex.common;
+
+public class StackElement {
+
+    int last;
+    int max;
+    boolean explore;
```
maronesy: As mentioned earlier nondexStart was not working.  I will test it with the new changes made.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,129 @@
...
+                    bufferedWriter.write(lastAndMax);
+                    bufferedWriter.newLine();
+                }
+                bufferedWriter.close();
+                return;
+            }
+            for (ExplorationEntry ch: choices) {
+                exploreStorage.add(ch.getShouldExplore());
+            }
+            if (!exploreStorage.contains(true)) {
```
maronesy: Fixed.

maronesy:
Thanks DrRacket.  I have fixed the issue in endRun.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: the ArrayList has been changed to a boolean.

nondex-common/src/main/java/edu/illinois/nondex/common/StackElement.java
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,142 @@
...
+            StackElement lastMax = choice.pop();
+            int last = lastMax.getLast();
+            int max = lastMax.getMax();
+            boolean explore = false;
+            if (last < max - 1) {
+                last++;
+                if (choice.size() > STARTING_COUNT) {
+                    explore = true;
+                }
+                StackElement lm = new StackElement(last, max, explore);
```
maronesy: now we are using the poped element.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,144 @@
...
+import java.util.List;
+import java.util.Random;
+import java.util.Stack;
+import java.util.logging.Level;
+
+public class SystematicRandom extends Random {
+    private final Stack<ExplorationEntry> choices;
+    private final String logFileName;
+    private final Configuration config = Configuration.parseArgs();
+    private List<String> lines;
```
alexgyori: this seems to me it should be a local variable in the constructor.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,144 @@
...
+    private int replayIndex;
+
+    public SystematicRandom() {
+        logFileName = config.systematicLog;
```
alexgyori: prefix fields with this. I like having them this way, because then I know on every access if it's an access local to the method or it's some object state.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,144 @@
...
+        choices = new Stack<ExplorationEntry>();
+        File file = new File(logFileName);
+        if (file.exists()) {
+            try {
+                lines = Files.readAllLines(Paths.get(logFileName));
+            } catch (IOException ioe) {
+                Logger.getGlobal().log(Level.SEVERE,"Could not read lines from systematic.log" ,ioe);
+            }
+            Object[] choiceValues;
```
alexgyori: Can this be String[]?

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,144 @@
...
+            for (String element: lines) {
+                String delimiter = "[ ]+";
+                if (!element.isEmpty()) {
+                    choiceValues = element.split(delimiter);
+                    if (choiceValues.length == 3) {
+                        int current = Integer.parseInt(choiceValues[0].toString());
```
alexgyori: Since your array will be String[] you will not need to call toString(). Similar on maximum, and shouldExplore.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,144 @@
...
+                        int maximum = Integer.parseInt(choiceValues[1].toString());
+                        boolean shouldExplore = Boolean.parseBoolean(choiceValues[2].toString());
+                        ExplorationEntry currentMaximum = new ExplorationEntry(current, maximum, shouldExplore);
+                        choices.push(currentMaximum);
```
alexgyori: merge the two statements choices.push(new Expl...);

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,144 @@
...
+                    } else {
+                        Logger.getGlobal().log(Level.SEVERE, "The 3 ExplorationEntry variable were not stored properly");
+                    }
+
+                }
+            }
+        }
+    }
+
+    public int nextInt(final int maximum) {
```
alexgyori: Add the following test:
x = nextInt(3);
if (x == 1)
  y = nextInt(2)
else
  y = nextInt(3)
print x,y

Your exploration should explore:
0,0
0,1
0,1
1,0
1,1
2,0
2,1
2,2

I believe you will explore 1,2 also, which is incorrect. Write a test and run it to confirm. Get back here if that's the case and we can think how to fix it. Otherwise I may be misunderstanding.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,144 @@
...
+            int current = currentMaximum.getCurrent();
+            int maximum = currentMaximum.getMaximum();
+            boolean shouldExplore = false;
+            if (current < maximum - 1) {
+                current++;
+                if (choices.size() > config.start) {
+                    shouldExplore = true;
+                }
+                currentMaximum.setCurrent(current);
+                currentMaximum.setShouldExplore(shouldExplore);
```
alexgyori: Something is fishy with the shouldExplore field of the entry. You are only writing it, and never using it in your logic that returns a random number. I think you should use it there, or if you think you shouldn't please remove it from everywhere.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,144 @@
...
+                        bufferedWriter.write(lastAndMax);
+                        bufferedWriter.newLine();
+                    }
+                } catch (IOException ioe) {
+                    Logger.getGlobal().log(Level.SEVERE,"Could not write systematic.log file" ,ioe);
+                }
+                return;
+            }
+
+            boolean allFalse = true;
```
alexgyori: why is this here? What is this logic doing?

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
```diff
@@ -0,0 +1,144 @@
...
+                if (ch.getShouldExplore()) {
+                    allFalse = false;
+                }
+            }
+            if (allFalse) {
+                break;
+            }
+        }
+
+        if (Files.exists(Paths.get(logFileName))) {
```
alexgyori: I would not delete the log file. It may be good to keep it in for debugging purposes at least and see its contents, even after the exploration is done, to check that everything is at max. This could be an assertion in your test, that after the exploration is finished, on the stack, current is equal to max in all positions where shouldExplore is true.

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
alexgyori: Okay.

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
```diff
@@ -94,6 +94,7 @@ public static Configuration getConfiguration() {
         public void run() {
             nondex.getConfig().createNondexDirIfNeeded();
             try {
+                nondex.getSysRandom().endRun();
```
alexgyori: do this only if the mode is systematic.

nondex-maven-plugin/src/it/systematic-it/SystematicTest.java
```diff
@@ -0,0 +1,51 @@
...
+ * Unit test for simple App.
+ */
+public class SysteamticTest {
+
+    @Test
+    public void test() {
+        Set<Integer> s = new HashSet<Integer>();
+        s.add(2);
+        s.add(3);
+	s.add(4);
```
alexgyori: This is misindented. Why did this pass Checkstyle? @DrRacket do you see why?

nondex-maven-plugin/src/it/systematic-it/invoker.properties
```diff
@@ -0,0 +1 @@
+invoker.goals = nondex:nondex
```
alexgyori: add newline at end of file.

nondex-maven-plugin/src/it/systematic-it/verify.groovy
```diff
@@ -0,0 +1,3 @@
+File nondexDirectory = new File( basedir, ".nondex" );
+
+assert nondexDirectory.isDirectory();
```
alexgyori: Add the assertion I mentioned earlier, asserting that at the end of exploration, the current equals max on all positions where the shouldExplore is true.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
```diff
@@ -59,33 +59,35 @@ a copy of this software and associated documentation files (the
...
-                            Pattern.compile(this.filter), this.start, this.end,
-                            Paths.get(this.baseDir.getAbsolutePath(), ConfigurationDefaults.DEFAULT_NONDEX_DIR).toString(),
-                            Paths.get(this.baseDir.getAbsolutePath(), ConfigurationDefaults.DEFAULT_NONDEX_JAR_DIR)
-                                .toString(),
-                            this.surefire, this.originalArgLine, this.mavenProject,
-                            this.mavenSession, this.pluginManager);
-            this.executions.add(execution);
-            allExceptions = this.executeSurefireExecution(allExceptions, execution);
-            this.writeCurrentRunInfo(execution);
+        if (this.mode.toString().equals("SYSTEMATIC")) {
```
alexgyori: @maronesy please address this.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
```diff
@@ -59,33 +59,33 @@ a copy of this software and associated documentation files (the
 public class NonDexMojo extends AbstractNonDexMojo {
 
     private List<NonDexSurefireExecution> executions = new LinkedList<>();
+    private MojoExecutionException allExceptions = null;
```
alexgyori: Why did you move this to a field? It doesn't seem to be something that needs to reside in a field.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
```diff
@@ -59,33 +59,33 @@ a copy of this software and associated documentation files (the
...
-                            this.surefire, this.originalArgLine, this.mavenProject,
-                            this.mavenSession, this.pluginManager);
-            this.executions.add(execution);
-            allExceptions = this.executeSurefireExecution(allExceptions, execution);
-            this.writeCurrentRunInfo(execution);
+        if (this.mode.toString().equals("SYSTEMATIC")) {
+            // In certain environments calls made to nextInt() from NonDex itself start at 59, if the user wants to change
+            // this variable they can use the DnondexStart argument
+            if (this.start == 0) {
+                this.start = 59;
```
alexgyori: Add a comment here explaining why you change this to 59 so that readers know in the future.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
```diff
@@ -59,33 +59,33 @@ a copy of this software and associated documentation files (the
...
+            }
+            do {
+                executeHelper(1);
+            } while (Files.exists(Paths.get(ConfigurationDefaults.DEFAULT_SYSTEMATIC_LOG)));
```
alexgyori: Oh, I see why you are removing it, because that's your stopping condition... You'll need to change this to communicate in some other way that exploration is finished. For example, by creating a file named explorationDone in which you store the results. On second thought, you could even simply rename the systematicLog file to systematicDone; this way your check will still work here, but you can still write the assertion I mentioned before and the log is still available for debugging.

nondex-common/src/main/java/edu/illinois/nondex/common/Configuration.java
```diff
@@ -178,7 +186,7 @@ public static Configuration parseArgs(Properties props) {
 
         final String testName = props.getProperty("test", null);
 
-        return new Configuration(nonDetKind, seed, filter, start, end, nondexDir, nondexJarDir, testName,
+        return new Configuration(nonDetKind, seed, filter, start, end, log, nondexDir, nondexJarDir, testName,
```
alexgyori: Add a method here getSystematicLogPath, that you write as the getFailuresPath is written, and use on all your accesses to the location of the log that method. That will ensure that the log file is created in the .nondex directory rather than in the project home somewhere.

nondex-maven-plugin/src/it/systematic-it/SystematicTest.java
DrRacket: @alexgyori No clear idea about this issue. I will look into it after all my finals..

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: Done.

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
maronesy: Done.

nondex-maven-plugin/src/it/systematic-it/SystematicTest.java
maronesy: This looks good when I commit it but once it gets pushed to repo it changes.

nondex-maven-plugin/src/it/systematic-it/invoker.properties
maronesy: Done.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
maronesy: renamed the file.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
maronesy: @DrRacket can you please elaborate do you mean not to use toString and instead use this.mode == ?

nondex-maven-plugin/src/it/systematic-it/SystematicTest.java
alexgyori: Just do not use tabs. Type spaces explicitly in your file. 
Your test is in the wrong place. Does this test ever run? It should be under src/it/systematic-it/src/main/java/.... Please see how the other integration tests are written. This is probably why Checkstyle doesn't check your file.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: so by default we are assuming that all elements are false unless we find that we still have an element with shouldExplore of true, if we do than we set allFalse to false and continue our exploration.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
maronesy: I have modified the code to rename the file.

nondex-common/src/main/java/edu/illinois/nondex/common/SystematicRandom.java
alexgyori: But shouldn't you use it also in nextInt? To decide if you need to explore or not?

alexgyori: Ok.

nondex-maven-plugin/src/it/systematic-it/SystematicTest.java
maronesy: I have moved it to the correct directly, mvn install/verify freezes now.  I believe it is because what I have added to pom:
<nondexMode> SYSTEMATIC </nondexMode>
When I remove this the test runs but it fails since it is in Full mode.

alexgyori: Did you try to run the tool in the Systematic mode on a maven project before? Are you saying that on the same code, if you run yourself mvn nondex:nondex with SYSTEMATIC it finishes, but mvn verify freezes?

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
alexgyori: He means using "this.mode == Mode.SYSTEMATIC". Please change this everywhere.

maronesy: I tried this but the build fails, I will try it again and paste the error here.

nondex-maven-plugin/src/it/systematic-it/pom.xml
```diff
@@ -0,0 +1,47 @@
...
+        <groupId>@project.groupId@</groupId>
+        <artifactId>@project.artifactId@</artifactId>
+        <version>@project.version@</version>
+      </plugin>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-surefire-plugin</artifactId>
+        <version>2.19.1</version>
+      </plugin>
+      <plugin>
```
alexgyori: you do not need to add this explicit dependency here. Did you look into what that @project.groupId@ does? That's a more robust way to specify this dependency.

PR #108, opened 2016-12-03
Modify JVMShutdownHook to throw RuntimeException when IOException occurs

DrRacket:
@alexgyori 
Since we are overriding a method with no "throws" signature and IOException is not a runtime exception, so I choose to throw a RunTimeException.
And I am kind of confused why there are two catch blocks in the original code. Code in the try block will only throw IOException, I don't see any chance that it will catch any other exceptions.


nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
```diff
@@ -108,10 +108,7 @@ public void run() {
             } catch (IOException ioe) {
                 Logger.getGlobal().log(Level.SEVERE,
                         "IOException when printing shuffling counts in shutdown hook.", ioe);
-            } catch (Throwable ex) {
```
alexgyori: Do not remove this code. This catch clause will catch and log errors at runtime, e.g., NullPointerException and other RuntimeExceptions that need not be declared and checked and log it properly.

PR #110, opened 2016-12-09
testing code for instrumentation of HashedMap in commons collections

whao2:
Instrumentation currently does not work. When running the tests for HashedMap after instrumentation, it complains about not being able to find the class any more.


alexgyori:
The build fails because you have your local path hard coded in the PR. Look at line: https://travis-ci.org/TestingResearchIllinois/NonDex/jobs/182675313#L1952

nondex-test/pom.xml
```diff
@@ -108,5 +108,12 @@
       <artifactId>nondex-common</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-collections4</artifactId>
+      <version>4.1</version>
+      <scope>system</scope>
+      <systemPath>/Users/kaku/Desktop/commons-collections4-4.1.jar</systemPath>
```
alexgyori: do not add this line here. This causes maven to not download the dependency into your local mvn repo.

alexgyori: If you do not give a system path, it will download the version you specify. In this case your version is 4.1, so it will download that. If you want 4.4.1 you should specify that version.

Issue #85, opened 2016-10-07
[BUG] NonDex does not run tests in clean


alexgyori:
NonDex does not run tests in the dry run.
The 1.1.1 release runs the tests in clean, but the current head doesn't.

If I remove the call to the method adding the exclude groups to the config, then everything is back to normal and tests run also in clean.

alexgyori:
@azy2 can you take a look and see if you spot some obvious issue?

azy2:
#113

PR #113, opened 2016-12-21
#85 fix

azy2:
Tests now run in clean with these changes but the build fails on comprehensive it. I'm not sure if this is because it's failing on the clean test or if we're not ignoring the clean failures appropriately. I don't know how comprehensive it is set up. @alexgyori can you take a look at this?


alexgyori:
The comprehensive is failing now because the tests are not running with NonDex (they're running only in the phase without nondex). The build fails because there's no test to run in nondex (surefire fails if there are no tests, so that's what the failure is about).
nondex-maven-plugin/target/it/comprehensive/build.log 

To convince myself that the failures in the dry-run are ignored I looked into the build.log of clean-it. That one has tests that fail in clean, but do not fail when run with NonDex. The result of the build there is success.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexSurefireExecution.java
```diff
@@ -91,6 +91,34 @@ protected void setupArgline() {
 
     }
 
+    @Override
+    protected Xpp3Dom applyNonDexConfig(Xpp3Dom configuration) {
+        return addExcludedGroups(super.applyNonDexConfig(configuration));
+    }
+
+    private Xpp3Dom addExcludedGroups(Xpp3Dom configNode) {
```
alexgyori: Why are we moving this to NonDex Surefire Execution? I thought we should apply the ignore everywhere.
This moving didn't fix the issue - it just changed so that tests were now not run with nondex while they were run in clean. Are you sure this needs to be moved?

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexSurefireExecution.java
```diff
@@ -91,6 +91,34 @@ protected void setupArgline() {
...
+        for (Xpp3Dom config : configNode.getChildren()) {
+            if ("excludedGroups".equals(config.getName())) {
+                Logger.getGlobal().log(Level.INFO, "Adding excluded groups to existing ones");
+                String current = config.getValue();
+                current = "," + current;
+                // It seems there is an error if you have the variable
```
alexgyori: Please add an integration test consisting of a multi module maven project; the project should define excluded groups both in the parent pom, and in the submodules. The tests should fail if they are not properly excluded. Then add also some tests in each module that have our NonDex Ignore. I think that because I remove the ${excludeGroups} I may be breaking multimodule projects, if the excluded groups defined in the parent pom are inherited through the variable I am removing.

nondex-maven-plugin/src/it/comprehensive/verify.groovy
```diff
@@ -4,3 +4,14 @@ assert nondexDirectory.isDirectory();
 
 // Every run adds 5 directories. Latest and jar are static
 assert (nondexDirectory.list().length - 2) % 5 == 0;
+
```
alexgyori: this code just makes sure the tests fail in clean, meaning they are run.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexSurefireExecution.java
azy2: I'm trying to employ polymorphism to only add the excludeGroups config to nondex executions. For some reason having the excludeGroups in a clean run stops the clean run from running at all.  In my testing NonDex still runs tests.

alexgyori: That is not what was stopping their execution. I think my current fix should fix also their running in the clean. Can you test that? Let me know if my change fixes also the clean run.

alexgyori: Also, if you want to test with the project that was failing, you can find that project at: nondex-maven-plugin/target/it/comprehensive/ (after running mvn verify -- it will generate there the concrete pom file and all the symbols from the .../src/it/comprehensive should be resolved.
You can there run mvn nondex:nondex with flags and what not and see the error messages.

azy2:
I tried adding a multi-module integration test per your instruction but I'm running into problems. I can't get it to build with multiple modules. If you take checkout the last commit and run mvn verify it builds and you'll notice there is another IT. But if you uncomment the modules part of nondex-maven-plugin/src/it/excluded-groups-it/pom.xml then it fails to build with a pretty cryptic error message. Can you take a look at that? I don't know enough about maven to figure out what's going on.

alexgyori:
@DrRacket can you look at this failures (as you have a mac)? This passes on linux, and I'm not sure what the issue is with the mac, as I can't even easily see the output log.

alexgyori:
@azy2 please add also some tests in that multi-module project. I'm not sure why it's failing on mac, but it maybe because there are no tests to run?

azy2:
I've tried adding tests but it seems as if maven is not running any tests at all. Even tests I add to the main module. I can't figure out why.

azy2:
I'm getting a build error on excluded-groups-it `No plugin found for prefix 'nondex' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/ben/research/NonDex/nondex-maven-plugin/target/local-repo), local.central file:///home/ben/.m2/repository), central (https://repo.maven.apache.org/maven2)] -> [Help 1]` @alexgyori  can you look into this? I don't know what it means. I have nondex installed on my machine.

nondex-maven-plugin/src/it/excluded-groups-it/pom.xml
```diff
@@ -18,24 +18,6 @@
       <module>module2</module>
     </modules>
 
-    <build>
```
alexgyori: @azy2 why did you remove this? Because you remove this, the build is broken. If you put it back, you will get some other issue, because module 2 has failing tests with nondex. Please debug that more.

PR #116, opened 2017-07-20
Fix NonDex for unresolved properties

alexgyori:
If the project has a property in argLine that is not defined, then the way NonDex invokes Surefire will cause it to crash.
This commit fixes it such that it will remove unresolved variables but leave the ones that are resolved in place.


nondex-maven-plugin/src/it/inexistent-variable-argline-it/pom.xml
```diff
@@ -0,0 +1,44 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
owolabileg: inexistent --> nonexistent in the IT dirname

nondex-maven-plugin/src/it/inexistent-variable-argline-it/verify.groovy
```diff
@@ -0,0 +1,3 @@
+File nondexDirectory = new File( basedir, ".nondex" );
+
+assert ! nondexDirectory.exists();
```
owolabileg: Does this verify.groovy check for nonexistent variable in `argLine`? I don't see how

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/CleanSurefireExecution.java
```diff
@@ -184,4 +186,15 @@ protected Xpp3Dom makeNode(String nodeName, String value) {
         node.setValue(value);
         return node;
     }
+
+    protected static String sanitizeAndRemoveEnvironmentVars(String toSanitize) {
```
owolabileg: Add a comment for what this method does. Is it merely removing illegal characters from var names?

nondex-maven-plugin/src/test/java/edu/illinois/nondex/plugin/CleanSurefireExecutionTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import static org.junit.Assert.assertEquals;
+
+import org.junit.Test;
+
+public class CleanSurefireExecutionTest {
+
+    @Test
+    public void testReplacementWorks() {
+        String result = CleanSurefireExecution.sanitizeAndRemoveEnvironmentVars("${somestuff}");
```
owolabileg: Why should `${somestuff}` be replaced? When is a variable considered "unresolved"?

nondex-maven-plugin/src/it/inexistent-variable-argline-it/pom.xml
alexgyori: inexistent is a word in Merriam-Webster.

nondex-maven-plugin/src/it/inexistent-variable-argline-it/verify.groovy
alexgyori: No. The inexistent variables are defined in the pom.xml on line 29: https://github.com/TestingResearchIllinois/NonDex/pull/116/files/fe5de11b6de60e6775caa2decc48791b73db9e25#diff-963ead46e2ddc1dd168aeed8eeb712d6R29 and the test that checks that the properties that exist are preserved are at: https://github.com/TestingResearchIllinois/NonDex/pull/116/files/fe5de11b6de60e6775caa2decc48791b73db9e25#diff-01bdd278b3993a77e1b8f4e0f3ad386aR13

nondex-maven-plugin/src/test/java/edu/illinois/nondex/plugin/CleanSurefireExecutionTest.java
alexgyori: a variable is unresolved if when I get the argLine string from Maven, it is not replaced.
Maven will replace all variables that are defined, but it just so happens in the project you mentioned that one of the variables is undefined, and then Maven will just let it go through unchanged; this breaks my code, so I am cleaning the string if I find any unresolved at this point because by  the time the sanitizer gets invoked there should be no unresolved vars.

nondex-maven-plugin/src/it/inexistent-variable-argline-it/verify.groovy
owolabileg: if that test fails, would this `verify.groovy` still pass the IT?

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/CleanSurefireExecution.java
alexgyori: No. The tests demonstrate what it does. It removes any string of the form ${something} which is how properties are referenced typically in maven. This is somewhat fishy that I can get unresolved variables from Maven, but I need to make my code robust.

nondex-maven-plugin/src/test/java/edu/illinois/nondex/plugin/CleanSurefireExecutionTest.java
owolabileg: Please add this as a code comment.

nondex-maven-plugin/src/it/inexistent-variable-argline-it/verify.groovy
alexgyori: No. Because by default the IT expects that the result of the build is PASS. If whatever the invoker.properties specifies should be ran fails (or more generally does not have the expected outcome because the expected outcome could be that the command should result in failure) then verify.groovy is not even invoked.

nondex-maven-plugin/src/test/java/edu/illinois/nondex/plugin/CleanSurefireExecutionTest.java
alexgyori: wilco

alexgyori: Done.

PR #118, opened 2017-07-25
decrease flakyness

alexgyori:
added a sleep in the IT verifier script because Windows' file system is sometimes slow and ITs can fail if the verify script starts too quickly. This shows up on AppVeyor.


owolabileg:
How does the flakiness here manifest, and how does `sleep(3000)` reduce  the flakiness. My understanding is that these `verify.groovy` scripts are invoked after the ITs complete their execution?

alexgyori:
They are invoked after the ITs finish the execution, but it somehow happens that the changes are not yet visible on the file system for some reason; waiting could probably be less than 3s.

This issue only appears on windows.

Issue #81, opened 2016-09-16
Help text for modes out of date


august782:
The current text for nondexMode when running "mvn nondex:help -Ddetail=true" describes four modes, which is out of date and not true. Please update.

august782:
Fixed by PR #122

PR #122, opened 2019-08-29
Fix the out-of-date help text for mode of debug in the maven plugin

Pezhin:
This fixes issue #81 as https://github.com/TestingResearchIllinois/NonDex/blob/a1d2fa21e19acd5453abf4c549b61101ac115e3b/nondex-common/src/main/java/edu/illinois/nondex/common/Mode.java#L32 indicates, there are only two modes: ONE and FULL. This PR removes the incorrect part.

Please note in order for the change to show up, you need to use the snapshot version when using the plugin.


Pezhin:
CI fails because `jdk_switcher: command not found`

It seems there is a bug in the CI script

PR #128, opened 2019-09-03
Updating checkstyle due to security vulnerabilities

august782:
Updating the checkstyle version to 8.18 due to security vulnerabilities. Additionally, updating the configuration files to match the new updated version and reformatting some source files that now fail some checkstyle checks.


Pezhin:
LGTM

Issue #123, opened 2019-08-30
"debug" mode throws NPE using the latest version


Pezhin:
Step to reproduce:
  1. `mvn nondex:clean`
  2. `mvn nondex:nondex`
  3. `mvn nondex:debug`

If using `1.1.1` version, `debug` should succeed and it generates a `debug` file containing the stacktrace. 
However, using `1.1.2-SNAPSHOT`, which is the build from cloning this repo, `debug` command crashes with NPE and does not generate `debug` file.

StackTrace
```
Caused by: java.lang.NullPointerException
    at edu.illinois.nondex.common.Configuration.getInvocationCount (Configuration.java:245)
    at edu.illinois.nondex.plugin.DebugTask.startDebugBinary (DebugTask.java:158)
    at edu.illinois.nondex.plugin.DebugTask.debugWithConfigurations (DebugTask.java:136)
    at edu.illinois.nondex.plugin.DebugTask.tryDebugSeeds (DebugTask.java:95)
    at edu.illinois.nondex.plugin.DebugTask.debug (DebugTask.java:85)
    at edu.illinois.nondex.plugin.DebugMojo.execute (DebugMojo.java:77)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo (DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:210)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
    at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
    at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
    at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
    at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
[...]

PR #130, opened 2019-09-08
Resetting the Surefire configuration between runs

august782:
In Issue #123, debug mode throws an NPE when debugging multiple tests because, upon reaching to debug the second test, the configuration is still stuck with using the execution ID from debugging the prior test.

The reason is that in between runs of NonDex the updated Surefire configuration (which passes in the relevant configuration options as arguments for running) is not reset, and it maintains some configuration options from before. The later runs, when parsing the arguments, always parses the old ones. This pull request resets the Surefire configuration between runs.

The other change, to NonDexSurefireExecution.java, is a by-product of doing the resetting, because integration tests now fail due to having configuration nodes without a <test> child. The change there simply adds one if it does not exist beforehand.

### DIFF
--- a/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/CleanSurefireExecution.java
+++ b/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/CleanSurefireExecution.java
@@ -87,6 +87,10 @@ public class CleanSurefireExecution {
     }
 
     public void run() throws MojoExecutionException {
+        Xpp3Dom origNode = null;
+        if (this.surefire.getConfiguration() != null) {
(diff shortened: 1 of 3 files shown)


Pezhin:
#123 seems to be fixed by this PR. 

The second commit aims to fix the issue with appveyor, right? The test still failed https://ci.appveyor.com/project/august782/nondex/builds/27257017#L2375. Did the second commit trigger Appveyor?

august782:
I could not figure out why the resetting of Surefire leads to the integration tests to fail in the Windows environment but not in the Linux one. It seems in the Windows environment not all directories are made in the end, yet the Linux one does. It's very interesting, but I don't even know how to begin debugging this. As such, for now, I just introduced an option to reset Surefire only when debugging.

PR #131, opened 2019-09-17
Removing shuffling for annotations

august782:
After re-examining the documentation related to getting annotations, it seems there is no language there that specifically states that the order is not guaranteed. As such, it may not be appropriate to shuffle the annotation order. This pull request simply removes the shuffling for uses of getting annotations.

If you know of where we initially found that annotation order is not guaranteed, please point it out and then we can keep the shuffling for the annotation order.

### DIFF
--- a/nondex-core/src/main/java/java/lang/Class.java
+++ b/nondex-core/src/main/java/java/lang/Class.java
@@ -3338,7 +3338,7 @@ public final class Class<T> implements java.io.Serializable,
      * @since 1.5
      */
     public Annotation[] getAnnotations() {
-        return edu.illinois.nondex.shuffling.ControlNondeterminism.shuffle(AnnotationParser.toArray(annotationData().annotations));
+        return AnnotationParser.toArray(annotationData().annotations);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


owolabileg:
Is shuffling of arrays returned by `get*Annotations()` causing some problem that you observed?

The specification of these methods is completely silent about the order of elements in the arrays, hence the reason why we considered it underdetermined in the first place. One could argue both ways.

august782:
It's not causing problems per se, but we get some tests flagged by NonDex due to annotation shuffling and want to be completely sure that it is actually a bug that's worth reporting. To be on the safe side, I think it would be best to only report those where we can confidently say order is not guaranteed based on the specification.

owolabileg:
So some tests assume that the order of annotations returned is fixed, and fail when NonDex shuffles the annotations?

august782:
Essentially, yes.

august782:
Latest commit was due to realizing that all I did was modify the committed JDK files instead of modifying the instrumentation code, so had to modify those and update tests so they don't expect annotation shuffling.

PR #140, opened 2020-04-09
Removing lengths list due to potential concurrency problems

august782:
This pull request removes the `lengths` list field, because it is not actually used elsewhere in the code and also can cause potential concurrency problems when NonDex is run in some projects.


Pezhin:
LGTM

Issue #89, opened 2016-10-08
Surefire missing NPE


alexgyori:
Fix NPE to display better error when the surefire plugin is missing from the pom

PR #154, opened 2020-11-24
Fix Surefire missing NPE Issue

ShawnLi-25:
#89 
* Handle NullPointerException error when surefire not explicitly declared in pom.xml
* Use surefire fallback version 3.0.0-M5
* Provide better error message

### DIFF
--- a/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/AbstractNonDexMojo.java
+++ b/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/AbstractNonDexMojo.java
@@ -181,7 +181,9 @@ public abstract class AbstractNonDexMojo extends AbstractMojo {
         this.surefire = this.lookupPlugin("org.apache.maven.plugins:maven-surefire-plugin");
 
         if (this.surefire == null) {
-            Logger.getGlobal().log(Level.SEVERE, "Make sure surefire is in your pom.xml");
+            Logger.getGlobal().log(Level.SEVERE, "Surefire is not explicitly declared in your pom.xml;"
+                    + " we will use version 3.0.0-M5, but you may want to change that.");
+            this.surefire = getSureFirePlugin();
         }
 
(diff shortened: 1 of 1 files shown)


nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/AbstractNonDexMojo.java
```diff
@@ -181,7 +181,9 @@ public void execute() throws MojoExecutionException, MojoFailureException {
         this.surefire = this.lookupPlugin("org.apache.maven.plugins:maven-surefire-plugin");
 
         if (this.surefire == null) {
-            Logger.getGlobal().log(Level.SEVERE, "Make sure surefire is in your pom.xml");
+            Logger.getGlobal().log(Level.SEVERE, "Surefire is not explicitly declared in your pom.xml;"
+                    + " we will use version 2.20, but you may want to change that.");
```
Anjiang-Wei, line 185: Surefire version seems to be updated to 2.22.2 instead of 2.20 as for NonDex itself.
Here I suppose the surefire version 2.20 is for the project NonDex intends to run upon. I am not sure whether 2.20 is the best choice.

It seems that a more recent surefire version (e.g., 3.0.0-M4) will be better, for both NonDex itself and the project under test, because it claims to provide better support for JUnit5. Please refer to https://maven.apache.org/surefire/maven-surefire-plugin/

ShawnLi-25: Sure. I agree that upgrade `Surefire` for all modules or `NonDex` and even the running project to the most recent surefire version (e.g.3.0.0-M5) is better.

Anjiang-Wei: Thanks for your change:)

darko-marinov:
Kaiyao confirmed via email that this functionality is included in the accepted big PR.

PR #157, opened 2020-12-02
Rewrite custom Level class in place of java.util.logging.Level

ShawnLi-25:
In Java 9+, due to the introduction of the [module system](https://openjdk.java.net/jeps/261), the original logger `java.util.logging` has been separated into an isolated module called `java.logging`. Our experiment indicates that `java.base` can not access any other module or jar during the initialization of boot layer even if we used `—add-reads` option.

We think the usage of class `Level` is minimal so we propose to reimplement the simplified version for the custom `Logger` instead of calling `java.util.logging.Level`.

P.S. Another solution can be using [`java.lang.System.Logger`](https://docs.oracle.com/javase/9/docs/api/java/lang/System.Logger.html) which is designed to log bootstrap information in Java9+.


august782 (approved):
This change looks fine to me. I can merge this part if there are no objections from others.

PR #159, opened 2020-12-02
Refactor Instrumenter to add Java9+ branching

ShawnLi-25:
* Modified `Instrumenter.java` to support Java9+ version. After Java9, `rt.jar` should be replaced by JRT filesystem to extract runtime classes to be instrumented.
* Add condition check to postpone `NonDex` object initialization in `ControlNondeterminism` after VM is booted in Java9+. (Java9+ changes the implementation in VM.class that loads the instrumented classes `java.util.HashMap/concurrent.ConcurrentHashMap`, which is instrumented by `NonDex`, the accidental order of class loading will crash the JVM and caused `ExceptionInInitializerError` error)
* Add Maven profiles to use different command arglines for Java8/Java9+
* We also did some refactoring work to eliminate duplicate code.

### DIFF
--- a/nondex-common/pom.xml
+++ b/nondex-common/pom.xml
@@ -45,5 +45,35 @@
       </plugin>
     </plugins>
   </build>
-  
+
+  <profiles>
+    <profile>
+      <id>java9+</id>
+      <activation>
+        <jdk>[9,)</jdk>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-javadoc-plugin</artifactId>
+            <configuration>
+              <additionalOptions>--add-exports java.base/jdk.internal.misc=ALL-UNNAMED</additionalOptions>
+            </configuration>
+          </plugin>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-compiler-plugin</artifactId>
+            <configuration>
+              <compilerArgs>
+                <arg>--add-exports</arg>
+                <arg>java.base/jdk.internal.misc=ALL-UNNAMED</arg>
+              </compilerArgs>
+            </configuration>
+          </plugin>
+        </plugins>
+      </build>
+    </profile>
+  </profiles>
+
 </project>
--- a/nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
+++ b/nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
@@ -29,16 +29,17 @@ WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
 package edu.illinois.nondex.shuffling;
 
 import java.io.IOException;
+import java.lang.reflect.Method;
 import java.nio.file.Files;
 import java.nio.file.StandardOpenOption;
 import java.util.Arrays;
-import java.util.LinkedList;
 import java.util.List;
 
 import edu.illinois.nondex.common.Configuration;
 import edu.illinois.nondex.common.Level;
 import edu.illinois.nondex.common.Logger;
 import edu.illinois.nondex.common.NonDex;
+import edu.illinois.nondex.common.Utils;
 
 public class ControlNondeterminism {
 
@@ -46,10 +47,12 @@ public class ControlNondeterminism {
 
     private static NonDex nondex;
 
+    private static boolean isCreatingNonDex = false;
+
     static {
         // Add shutdown hook
         Runtime.getRuntime().addShutdownHook(ControlNondeterminism.jvmShutdownHook);
-        nondex = new NonDex();
+        initializeNondex();
     }
 
     public static Configuration getConfiguration() {
@@ -60,6 +63,13 @@ public class ControlNondeterminism {
         if (originalOrder.size() < 2) {
             return originalOrder;
         }
+
+        initializeNondex();
+
+        if (nondex == null) {
+            return originalOrder;
+        }
+
         return nondex.getPermutation(originalOrder);
     }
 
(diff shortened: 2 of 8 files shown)


nondex-common/pom.xml
```diff
@@ -45,5 +45,35 @@
       </plugin>
     </plugins>
   </build>
-  
+
+  <profiles>
```
august782, line 49: So from my understanding, this profile is to ensure that if the underlying JVM is Java 9 or up, then the way to build the NonDex project should be configured differently? If NonDex is actually run on a project that uses Java 9+, would a NonDex built using Java 8 work on it?

ShawnLi-25: > So from my understanding, this profile is to ensure that if the underlying JVM is Java 9 or up, then the way to build the NonDex project should be configured differently? If NonDex is actually run on a project that uses Java 9+, would a NonDex built using Java 8 work on it?

Yes, you're correct. It is used to configure NonDex correctly to be able to build under Java 9 or up.

No, our current implementation requires the underlying JVM version to be consistent between `running project` and `NonDex` itself. Otherwise, there will be exceptions caused by incompatible version of instrumented classes. For E.g. if `NonDex` is built using Java 8 but the project is built with Java 11.
```
java/lang/UnsupportedClassVersionError: java/lang/Class has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
```diff
@@ -60,6 +63,13 @@ public static Configuration getConfiguration() {
         if (originalOrder.size() < 2) {
             return originalOrder;
         }
+
+        initializeNondex();
```
august782, line 67: Maybe put some comment to remind ourselves why we have to call initializeNondex here even though it is called in static block, that there is this issue with Java 9 classloading.

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
```diff
@@ -95,6 +111,33 @@ public static Configuration getConfiguration() {
...
+            if (nondex == null) {
+                nondex = NonDex.getInstance();
+            }
+            return;
+        }
+
+        try {
+            Method isBooted = Class.forName("jdk.internal.misc.VM").getDeclaredMethod("isBooted");
+            if ((Boolean)isBooted.invoke(null) && nondex == null && !isCreatingNonDex) {
+                isCreatingNonDex = true;
```
august782, line 131: Since you have to use this boolean isCreatingNonDex to set the nondex instance, are there some race conditions to be concerned about here? Are these booleans enough to prevent those? Also, is it necessary to set isCreatingNondex back to false afterwards?

nondex-common/src/main/java/edu/illinois/nondex/shuffling/ControlNondeterminism.java
```diff
@@ -95,6 +111,33 @@ public static Configuration getConfiguration() {
...
+                // Call getStackTrace here to bypass exception "can't initialize class StackTraceElement$HashedModules"
+                Thread.currentThread().getStackTrace();
+                nondex = NonDex.getInstance();
+            }
+        } catch (Exception ex) {
```
august782, line 136: Is there some more specific exception we can catch here as to ensure it's an issue with the reflection trying to load a class?

nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/Instrumenter.java
```diff
@@ -92,40 +107,94 @@ private Instrumenter() {
...
+            ZipFile rt = null;
+            try {
+                rt = new ZipFile(rtPath);
+            } catch (IOException exc) {
+                Logger.getGlobal().log(Level.SEVERE, "Are you sure you provided a valid path to your rt.jar?");
+                throw exc;
+            }
+            rtZipFile = rt;
+        } else {
+            rtFileSystem = FileSystems.getFileSystem(URI.create("jrt:/"));
```
august782, line 168: I assume this is the new part to consider for Java 9? Maybe add a comment to remind us about that.

nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/Instrumenter.java
```diff
@@ -92,40 +107,94 @@ private Instrumenter() {
...
+                if (entry == null) {
+                    Logger.getGlobal().log(Level.WARNING, "Could not find " + className + " in rt.jar");
+                    Logger.getGlobal().log(Level.WARNING, "Are you running java 8?");
+                    Logger.getGlobal().log(Level.WARNING, "Continuing without instrumenting: " + className);
+                    return null;
+                }
+                clInputStream = rtZipFile.getInputStream(entry);
+                return clInputStream;
+            } catch (IOException exc) {
+                Logger.getGlobal().log(Level.WARNING, "Cannot find " + className + " are you sure this is a valid rt.jar?");
```
august782, line 129: There used to be a comment here about not knowing whether this code is reachable. Do you think it is for sure reachable? Otherwise, we may want to just keep the TODO around until we get to it.

nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/Instrumenter.java
```diff
@@ -147,31 +216,17 @@ private void process(String rtJar, String outJar)
...
-                Logger.getGlobal().log(Level.WARNING, "Continuing without insturmenting: " + cl);
+            InputStream clInputStream = this.getClassInputStream(cl);
+            if (clInputStream == null) {
                 continue;
             }
 
             this.writeMd5(clInputStream, cl + ".md5", outZip);
 
-            clInputStream = rt.getInputStream(rt.getEntry(cl));
+            clInputStream = this.getClassInputStream(cl);
```
august782, line 229: Is it necessary to do this call again, seeing that it is assigned this way at the beginning of the loop?

PR #162, opened 2021-10-03
performance improvement on hashmap

Jiaming1999:
Using entrySet instead of keySet, we do not have to access hashmap again to obtain the value. This could be a potential performance improvement for many values that share one hash value.
- The fixing report is generated by Amazon CodeGuru reviewer
- Test cases in nondex-maven-plugin module passed


darko-marinov (approved):
Thanks for the PR. It marginally improves performance but marginally decreases readability.

PR #164, opened 2022-04-10
upgrade junit from 4.12 to 4.13.2

Instein98:


darko-marinov (approved):
Thanks for the contribution, Yicheng. Let me just document here that you added annotations more annotations to reduce the chance of the test to fail due to "bad" random choices.

PR #165, opened 2022-07-12
add group

MarcyGO:
add group


darko-marinov (approved):
This is a good starting point. By Monday, let's support (1) Java 9+ and (2) passing at least one random seed.

PR #155, opened 2020-11-26
Introduce support for Java9/10/11

ShawnLi-25:
### This is the final cumulative PR for the whole feature. We also break it down into multiple small patches to be more readable for review.

Our work enables NonDex to support Java9/10/11 Maven projects. It's also compilable in either Java8 or Java9/10/11 environment. The result of our recent experiments on [open-source project dataset](https://docs.google.com/spreadsheets/d/1OnUjKjamB4HWS_eDG2CXctYq36xuP727xx0CalK-B7A/edit#gid=0) proves that the tool works fine on both Java8 and Java9/10/11 projects.

The migration job can be broke down as the following:
1. Use **JRT filesystem** to extract runtime classes to be instrumented in replace of `rt.jar` in Java9+ 
(Reason: JRE images in Java9+ no longer contain the files lib/rt.jar, lib/tools.jar, lib/dt.jar, and other internal JAR files.)
2. Implement custom **Level** class to replace `java.util.logging.Level`.
(Reason: In Java 9+, the original logger `java.util.logging` has been separated into an isolated module called `java.logging`, and we can not access the logger in `java.base` module in Java 9+).
3. Postpone `NonDex` initialization in `ControlNondeterminism` after VM is booted, otherwise we'll have `ExceptionInInitializerError` when instrumenting `java.util.HashMap/concurrent.ConcurrentHashMap` 
(Reason: Java9+ changes the implementation of `VM.class` that loads these class too early even during VM initPhase. Since `NonDex` instruments these classes, the accidental order of class loading will crash the JVM)
4. Refactor `Instrumenter.java` to be workable with Java9+.
5. Update related Maven plugins & Upgrade ASM5 to ASM9 to support later Java versions.

Please feel free to comment and follow up. Thank you!

PR #166, opened 2022-07-15
Add support for Java 9+: Addressed code changes in #155 and resolved conflicts.

kaiyaok2:
In recent tests with Java 9/10/11/17 with all code changes in https://github.com/TestingResearchIllinois/NonDex/pull/155, no problems are found. This PR addresses the changes in https://github.com/TestingResearchIllinois/NonDex/pull/155, and makes minor changes to resolve its conflicts with the current main branch.

After this PR NonDex Maven plugin (1.1.3-SNAPSHOT) shall work with both Java 8 and Java 9+.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -39,6 +39,9 @@ buildNumber.properties
 .idea/*
 *.iml
 
+### MAC ###
+.DS_Store
+
 ### Emacs ###
 # -*- mode: gitignore; -*-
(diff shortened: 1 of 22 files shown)


darko-marinov (approved):
Let's accept this as we discussed offline so we can move on. Please open smaller PRs.

PR #53, opened 2016-06-15
Debugging reports multiple

august782:
This change modifies debugging to try many (single) points of shuffling for all found failing seeds.

PR #168, opened 2022-08-09
Allow debug to report multiple invocations

kaiyaok2:
This PR addresses the idea in (https://github.com/TestingResearchIllinois/NonDex/pull/53) to report many invocations from all seeds that can reproduce error. 
3 changes are made w.r.t the original PR: 
- Resolve code conflict with the latest master branch
- If a test is ND (falsely identified as ID by NonDex, which is quite common), `reportDebugInfo()` , which subsequently calls `failsWithConfig()` can return `null` even if being run with the same "originally failing" configuration. Then, the list `allFailingConfigurations` can contain `null` entries, which causes NPE when iterating through the list. The solution is to add a null check to avoid NPE thrown by `makeResultString()`.
- The original PR did not add the results found by linear search to the output list. This PR corrects it.

### DIFF
--- a/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/DebugTask.java
+++ b/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/DebugTask.java
@@ -247,6 +247,11 @@ public class DebugTask {
                     Logger.getGlobal().log(Level.FINE, "Refining did not work. Does not fail with linear on range "
                         + localStart + " : " + localEnd + ".");
                 }
+            } else {
+                // Since start <= end is always true, this branch means start == end, so reached end
+                if (failingConfiguration != null) {
+                    allFailingConfigurations.add(this.reportDebugInfo(failingConfiguration));
+                }
             }
         }
         return allFailingConfigurations;
--- a/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/DebugTask.java
+++ b/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/DebugTask.java
@@ -37,6 +37,8 @@ import edu.illinois.nondex.common.Level;
 import edu.illinois.nondex.common.Logger;
 import edu.illinois.nondex.common.Utils;
 
+import org.apache.commons.lang3.tuple.Pair;
+
 import org.apache.maven.execution.MavenSession;
 import org.apache.maven.model.Plugin;
 import org.apache.maven.plugin.BuildPluginManager;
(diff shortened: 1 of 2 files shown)


nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/DebugTask.java
```diff
@@ -128,85 +147,114 @@ private List<Configuration> createNewSeedsToRetry() {
...
     }
 
-    public Configuration startDebugBinary(Configuration config) {
-        long start = 0;
-        long end = config.getInvocationCount();
+    public List<Configuration> startDebugBinary(Configuration config) {
+        List<Configuration> allFailingConfigurations = new LinkedList<Configuration>();
+
+        List<Pair<Pair<Long, Long>, Configuration>> pairs = new LinkedList<Pair<Pair<Long, Long>, Configuration>>();
+        pairs.add((Pair<Pair<Long, Long>, Configuration>)Pair.of((Pair<Long, Long>)Pair.of(0L,
```
august782, line 172: Perhaps the initial entry in the list should map the start/end points to the `config`, as it seems the invariant should be that the start/end points map to a failing configuration.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/DebugTask.java
```diff
@@ -128,85 +147,114 @@ private List<Configuration> createNewSeedsToRetry() {
...
     private Configuration reportDebugInfo(Configuration failingConfiguration) {
         return this.failsWithConfig(failingConfiguration, failingConfiguration.start, failingConfiguration.end, true);
     }
 
-    public Configuration startDebugLinear(Configuration config, long start, long end) {
+    public List<Configuration> startDebugLinear(Configuration config, long start, long end) {
+        List<Configuration> allFailingConfigurations = new LinkedList<Configuration>();
+
+        List<Pair<Pair<Long, Long>, Configuration>> pairs = new LinkedList<Pair<Pair<Long, Long>, Configuration>>();
+        pairs.add((Pair<Pair<Long, Long>, Configuration>)Pair.of((Pair<Long, Long>)Pair.of(start, end),
```
august782, line 221: Same thing here about initial configuration in the list.

kaiyaok2: Thanks for pointing that out!

darko-marinov:
The code looks good enough to me to merge so we can release sooner the next version of NonDex to support Java 9+.

darko-marinov (approved):
The code looks good enough to me to merge so we can release sooner the next version of NonDex to support Java 9+.

PR #172, opened 2022-08-31
change default version

kaiyaok2:
Fix IT test failures


darko-marinov (approved):
Do we need to update something in the release nondex-2.0 because of this?

PR #173, opened 2022-08-31
Added filtering to auto-configure version

kaiyaok2:
Using maven filtering from source file to auto-configure version. Avoid needs to manually change versions in future upgrading.


nondex-common/pom.xml
```diff
@@ -13,6 +13,12 @@
 
   
   <build>
+    <resources>
+       <resource>
+	       <directory>src/main/java/edu/illinois/nondex/common</directory>
```
darko-marinov, line 18: wrong indentation

darko-marinov (changes requested):
Besides that space issue, the rest looks fine if you tested it.

kaiyaok2:
Thanks for pointing that out! I resolved and squashed commits

Best,
Kaiyao

kaiyaok2:
Closing as the changes are addressed in https://github.com/TestingResearchIllinois/NonDex/pull/174

PR #176, opened 2022-11-08
Ensure Sanity for Sonatype checks

kaiyaok2:
This PR addressed the code changes for sanity w.r.t. Maven Central Deployment. It is also the last PR for NonDex 2.1. It does the followings:
1. Version 1.6.7 of `nexus-staging-maven-plugin` is deprecated and causes failures in deploying. Bump the version to 1.6.13.
2. If we use the `mvn deploy -P` scheme (rather than `mvn release:prepare` and `mvn release:perform`), we need to also configure `org.sonatype.plugins` (rather than just `maven-deploy-plugin`) to skip deployment, or that the integration test module will still be deployed.
3. Sonatype enables a scheme to ensure a source code `.jar` file and a Javadoc `.jar` file presented in all submodules. However, as our test module has no `src/main` folder, such jars are not generated, and sanity check fails. The Sonatype maintainers suggest us to generate these empty jars manually, so we used `maven-jar-plugin` to do so.

### DIFF
--- a/nondex-test/pom.xml
+++ b/nondex-test/pom.xml
@@ -15,6 +15,9 @@
     <profile>
       <id>java8Below</id>
       <activation>
+        <property>
+          <name>!deploy</name>
+        </property>
         <jdk>(,1.8]</jdk>
       </activation>
       <build>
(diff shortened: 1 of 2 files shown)


darko-marinov (approved):
It worked for the release (although requires Java 11), so I'll approve all changes.

PR #182, opened 2023-09-21
[maven-release-plugin] prepare release nondex-2.1.1

kaiyaok2:
As title


darko-marinov:
The version 2.1.1 and 2.1 share the exact same source code, but 2.1,1 was compiled with Java 8, while 2.1 was complied with Java 11.

darko-marinov (approved):
Kaiyao and I are tagging the repo together in person. An issue was that 2.1 was mistakenly complied with Java 11, so 2.1.1 is the same source code compiled with Java 8.

PR #185, opened 2023-10-20
Refactored code to write run id to execution file

kavvya97:
**Description**
Every  `NonDexSurefireExecution` has an `executionId` and a `runFilePath`. The `runFilePath` is configured based on the `executionId` [link](https://github.com/TestingResearchIllinois/NonDex/blob/master/nondex-common/src/main/java/edu/illinois/nondex/common/Configuration.java#L213). The `NonDexMojo` class consists of a `List<NonDexSurefireExecution>` called `executions`. Within the `execute` method, this list is populated.

For each `execution` in this list , the `executionId` of the corresponding execution is being appended to a file whose path is the `runFilePath` of the first member of this list. This base path information is not stored anywhere and fetched from the first element for writing each `executionId`. This is the portion of the code being refactored as part of this Pull request.

**Fix**
The base path is being used across functions of the class and therefore, It is necessary to extract it as a member of the `NonDexMojo` class. This allows future configurations of this base path with a single change. This member is extracted within `execute` function and is used in the `writeCurrentRunInfo` method.

**Verification**
- The fix was verified by checking that a single `<executionId>.run` file was created within the `.nondex` directory, containing all the `executionIds` of the run, when we ran mvn test on a sample repository

`mvn -pl pushy edu.illinois:nondex-maven-plugin:2.1.7-SNAPSHOT:nondex -Dtest=com.eatthepath.pushy.apns.server.TokenAuthenticationValidatingPushNotificationHandlerTest#testHandleNotificationWithWithExpiredAuthenticationToken`

- The number of generated files were verified by [integration test](https://github.com/TestingResearchIllinois/NonDex/blob/master/nondex-maven-plugin/src/it/simple-it/verify.groovy#L7) in Nondex repository

### DIFF
--- a/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
+++ b/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
@@ -34,6 +34,7 @@ import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.nio.file.Files;
+import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.nio.file.StandardOpenOption;
@@ -65,6 +66,11 @@ public class NonDexMojo extends AbstractNonDexMojo {
     private List<NonDexSurefireExecution> executions = new LinkedList<>();
     private ArrayList<CleanSurefireExecution> executionsWithoutShuffling =
             new ArrayList<CleanSurefireExecution>();
+    private Path runFilePath = null;
+
+    public void setFilePath(Path value) {
+        this.runFilePath = this.runFilePath == null ? value : this.runFilePath;
+    }
 
     @Override
     public void execute() throws MojoExecutionException, MojoFailureException {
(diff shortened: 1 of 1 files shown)


darko-marinov:
Thanks for the PR! Why do we need this change? What problem does it solve?

kavvya97:
> Why do we need this change? What problem does it solve?

There is no change in functionality. It is just to fix the TODO and improve the code quality.

darko-marinov:
@kaiyaok2 can you please inspect this change when you find some time? Does this pass all the tests?

kaiyaok2:
@kavvya97 Thanks for your contribution! Tested it on a few repositories and the change looks solid.

darko-marinov:
Accepting per Kaiyao's suggestion.

nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/NonDexMojo.java
```diff
@@ -65,6 +66,11 @@ public class NonDexMojo extends AbstractNonDexMojo {
     private List<NonDexSurefireExecution> executions = new LinkedList<>();
     private ArrayList<CleanSurefireExecution> executionsWithoutShuffling =
             new ArrayList<CleanSurefireExecution>();
+    private Path runFilePath = null;
+
+    public void setFilePath(Path value) {
+        this.runFilePath = this.runFilePath == null ? value : this.runFilePath;
```
darko-marinov, line 72: This may have been easier to read with an `if`, and I'm not sure how this entire thing works with threads, but let me accept.

darko-marinov (approved):

PR #186, opened 2023-10-20
Fixed TODO to improve unordered string comparision

kavvya97:
### Setup
- Java version: 11.0.20.1
- Maven version: Apache Maven 3.6.3

### Description
This PR addresses the following TODO in the test utility function:

https://github.com/kavvya97/NonDex/blob/8843b436802af4a0850f63d6ae0f231164dafe78/nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java#L63

The function `assertEqualstUnordered` checks whether the contents of two data structures are equal without considering the order of the elements. Currently, it converts the content of both structures into string form and checks if the strings have the same length. If they do, it tokenizes one of them by splitting on commas to obtain multiple `key=value` elements. Then, it checks if each element is present in the other string as a substring.

https://github.com/kavvya97/NonDex/blob/8843b436802af4a0850f63d6ae0f231164dafe78/nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java#L66

However, this approach doesn't account for duplicates or substrings in the data structures. For example, consider the two strings:
`"{0=0, 0=0, 1=1, 2=2}"` and `"{0=0, 1=1, 2=2, 1=1}"`. These two strings should not be considered equal, but the current test considers them equal.

### Fix
Both strings are tokenized in the same manner as described above. The tokenized strings are then sorted and checked for equality. This approach will detect cases where duplicates differ between the two strings.

### Verification
- The fix did not introduce any regressions in existing tests that use this utility.
- An additional test was implemented to handle corner cases, such as the example mentioned above. The existing implementation of `assertEqualstUnordered` fails this test.


nondex-test/src/test/java/edu/illinois/nondex/core/HashMapTest.java
```diff
@@ -148,6 +148,17 @@ public void testEntrySet() {
                 "{0=0, 1=1, 2=2, 3=3, 4=4, 5=5, 6=6, 7=7, 8=8, 9=9}", entrySet.toString());
     }
 
+    @Test
```
darko-marinov, line 151: Please use `@Test(expected=....)` to simplify the test when it should pass when an exception is thrown.

nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java
```diff
@@ -56,16 +58,25 @@ protected void assertParameterized(T ds, Object derived, String str) {
...
+    }
 
+    protected void assertEqualstUnordered(String msg, String expected, String actual) {
+        String[] actualTokenized = this.formatString(actual);
+        String[] expectedTokenized = this.formatString(expected);
 
+        Arrays.sort(actualTokenized);
+        Arrays.sort(expectedTokenized);
+        if (Arrays.equals(actualTokenized, expectedTokenized)) {
+            Assert.assertTrue(msg + ": " + expectedTokenized + "=/= " + actualTokenized, true);
```
darko-marinov, line 77: `assertTrue(message, true)` can never fail. What is the goal of this assert?

darko-marinov: It seems you want to use `assertArraysEquals`.

nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java
```diff
@@ -56,16 +58,25 @@ protected void assertParameterized(T ds, Object derived, String str) {
...
         String[] elems = trimmed.split(",");
-        // TODO(gyori): fix and make this more robust. It does not check duplicates, substrings, etc.
         for (int i = 0; i < elems.length; i++) {
             elems[i] = elems[i].trim();
-            Assert.assertTrue(msg + ": " + trimmed + " =/= " + actual, actual.contains(elems[i]));
         }
+        return elems;
+    }
 
+    protected void assertEqualstUnordered(String msg, String expected, String actual) {
```
darko-marinov, line 70: Where is `assertEqualsUnordered` used?

darko-marinov (changes requested):
Where is this method used?

nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java
kavvya97: The class `HashMapTest` utilizes this function in the following test method
`smokeTest`
`testKeySet`
`testValues`

nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java
kavvya97: sure. we will incorportate this

nondex-test/src/test/java/edu/illinois/nondex/core/HashMapTest.java
kavvya97: Will incorporate these changes

nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java
```diff
@@ -56,16 +58,22 @@ protected void assertParameterized(T ds, Object derived, String str) {
...
+        String trimmed = expected.substring(1, expected.length() - 1);
+        String[] actualTokenized = this.formatString(actual);
```
darko-marinov, line 72: Is `formatString` the best name? The method seems to be splitting a collection into elements. Even the input is not an arbitrary `msg`.

darko-marinov (changes requested):
The assertions look much better now. Let's clean the names.

nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java
```diff
@@ -56,16 +58,22 @@ protected void assertParameterized(T ds, Object derived, String str) {
         }
     }
 
-    protected void assertEqualstUnordered(String msg, String expected, String actual) {
-        Assert.assertEquals(msg + ": " + expected + " =/= " + actual, expected.length(), actual.length());
-        String trimmed = expected.substring(1, expected.length() - 1);
+    private String[] trimAndSplitStrings(String msg) {
+        String trimmed = msg.substring(1, msg.length() - 1);
         String[] elems = trimmed.split(",");
```
darko-marinov, line 63: Can it `split(", ")` and then not even require the loop for `trim()`?

nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java
```diff
@@ -46,7 +48,7 @@ protected void assertParameterized(T ds, Object derived, String str) {
             case FULL:
                 String tempStr = derived.toString();
                 Assert.assertNotEquals("FULL is improperly running", str, tempStr);
-                this.assertEqualstUnordered("Does not match permutation", str, tempStr);
+                this.assertEqualsUnordered("Does not match permutation", str, tempStr);
```
darko-marinov, line 51: Any idea why they ever called it `Equalst`? Was it a misspelling or some joke about `EqualsString`? Does the Git log show anything useful?

nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java
kavvya97: yes that should work. Will update the PR

nondex-test/src/test/java/edu/illinois/nondex/core/AbstractCollectionTest.java
kavvya97: Hello Professor, I checked the git blame and from that I can infer that it was initially called as such. this was the first time when the function was commited/created
https://github.com/TestingResearchIllinois/NonDex/commit/d4f6e033687d0203a4f6557405345dc9422b240a#diff-b6d9573a5585dde09208a701b942a04b7e8017e4916381997baa33cf6d32edabR38

PR #188, opened 2023-10-30
Gradle script

hofi1:
Implements a script to modify the gradle files of your project


kaiyaok2:
@hofi1 Thanks!! The change looks good to me. I'm just wondering if we could remove Python as a dependency (i.e. let bash do all the work). Please review the script Darko shared and feel free to incorporate part of that to this PR

darko-marinov:
Thanks for the readme. It can remain the same even if we remove the dependency on Python and use only shell. Can you check Kaiyao's emails for shell only?

PR #193, opened 2024-10-15
Fix JVM crashing issues

kaiyaok2:
2 Changes:
- Do a merge when multiple `--patch-modules java.base` in argline
- Remove placeholders in final argline

### DIFF
--- a/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/CleanSurefireExecution.java
+++ b/nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/CleanSurefireExecution.java
@@ -102,6 +102,7 @@ public class CleanSurefireExecution {
                                    + " " + domNode + " "
                                    + MojoExecutor.executionEnvironment(this.mavenProject, this.mavenSession,
                                                                        this.pluginManager));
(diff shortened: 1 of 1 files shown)


nondex-maven-plugin/src/main/java/edu/illinois/nondex/plugin/CleanSurefireExecution.java
```diff
@@ -102,6 +102,34 @@ public void run() throws MojoExecutionException {
                                    + " " + domNode + " "
                                    + MojoExecutor.executionEnvironment(this.mavenProject, this.mavenSession,
                                                                        this.pluginManager));
+            Xpp3Dom domNodeArgLine = domNode.getChild("argLine");
```
darko-marinov, line 105: Can you add some comment for these changes? It's not clear why this is being done. Even better, you can extract this code into a method by itself, and then name it appropriately.

kaiyaok2: thanks! done.

darko-marinov (approved):
Thanks for promptly fixing this and for testing on some repos. We'll have students test on more repos.

PR #195, opened 2024-10-25
Version bump to 2.2.1-SNAPSHOT

kaiyaok2:


darko-marinov (approved):
I'll accept this commit with 7 files changed, but wouldn't it be better to put the version number in a variable in this top-level pom.xml and then use that variable in the other files?

PR #196, opened 2024-10-25
Support Java 21

kaiyaok2:


darko-marinov (approved):
Fantastic to get support for Java 21 with just these version changes. Let me copy from your email (because your PR had no text :):

> The work is as easy as the Java 11 -> 17 update, where I just need to bump ASM to the newer version that supports Java 21, and also update maven-jar-plugin, maven-shade-plugin and maven compiler plugin in a similar way.

PR #198, opened 2024-11-20
Add workaround for KeySpliterator shuffling

everbrightw:
- Introduced shuffling in the `KeySpliterator` of `HashMap`.

### Remaining Work
- [x] Add shuffling for `KeySpliterator`
- [ ] Add shuffling for `ValueSpliterator`
- [ ] Add shuffling for `EntrySpliterator`

### DIFF
--- a/nondex-core/src/main/java/java/util/HashMap.java
+++ b/nondex-core/src/main/java/java/util/HashMap.java
@@ -1517,9 +1517,13 @@ public class HashMap<K,V> extends AbstractMap<K,V>
     static final class KeySpliterator<K,V>
         extends HashMapSpliterator<K,V>
         implements Spliterator<K> {
+
+        private final KeySpliteratorShuffler<K, V> shuffler;
+
         KeySpliterator(HashMap<K,V> m, int origin, int fence, int est,
                        int expectedModCount) {
             super(m, origin, fence, est, expectedModCount);
+            this.shuffler = new KeySpliteratorShuffler<>(this);
         }
 
         public KeySpliterator<K,V> trySplit() {
@@ -1530,6 +1534,10 @@ public class HashMap<K,V> extends AbstractMap<K,V>
         }
 
         public void forEachRemaining(Consumer<? super K> action) {
+            shuffler.forEachRemaining(action);
+        }
+
+        public void original_forEachRemaining(Consumer<? super K> action) {
             int i, hi, mc;
             if (action == null)
                 throw new NullPointerException();
--- a/nondex-core/src/main/java/java/util/KeySpliteratorShuffler.java
+++ b/nondex-core/src/main/java/java/util/KeySpliteratorShuffler.java
@@ -1,0 +1,24 @@
+package java.util;
+import java.util.function.Consumer;
+import java.util.HashMap.KeySpliterator;
+
+public class KeySpliteratorShuffler<K, V> {
+
+    private final Iterator<K> iter;
+    private final KeySpliterator<K, V> keySpliterator;
+
+    public KeySpliteratorShuffler(KeySpliterator<K, V> ks) {
+        this.keySpliterator = ks;
+
+        final List<K> keys = new ArrayList<>();
+        keySpliterator.original_forEachRemaining(key -> keys.add(key));
+
+        List<K> res = edu.illinois.nondex.shuffling.ControlNondeterminism.shuffle(keys);
+
+        iter = res.iterator();
+    }
+
+    public void forEachRemaining(Consumer<? super K> action) {
+        iter.forEachRemaining(action);
+    }
+}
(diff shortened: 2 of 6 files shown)


nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/HashKeySpliteratorASMDump.java
```diff
@@ -0,0 +1,155 @@
...
+            methodVisitor.visitInsn(RETURN);
+            methodVisitor.visitMaxs(2, 0);
+            methodVisitor.visitEnd();
+        }
+        classWriter.visitEnd();
+
+        return classWriter.toByteArray();
+    }
+}
+
```
darko-marinov, line 155: Extra empty line at the end?

nondex-core/src/main/java/java/util/KeySpliteratorShuffler.java
```diff
@@ -0,0 +1,24 @@
...
+
+        List<K> res = edu.illinois.nondex.shuffling.ControlNondeterminism.shuffle(keys);
+
+        iter = res.iterator();
+    }
+
+    public void forEachRemaining(Consumer<? super K> action) {
+        iter.forEachRemaining(action);
+    }
+}
```
darko-marinov, line 24: Missing end-of-line at the end?

darko-marinov (changes requested):
I've added trivial comments while making a sanity pass.

@kaiyaok2 can you please review this PR in more detail?

everbrightw:
updated for resolving `checkstyle`

nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/Instrumenter.java
```diff
@@ -72,6 +73,7 @@ public final class Instrumenter {
     public static final String hashMapNodeName = "java/util/HashMap$Node.class";
     public static final String hashMapEntryName = "java/util/HashMap$Entry.class";
     public static final String hashMapHashIteratorShufflerName = "java/util/HashMap$HashIterator$HashIteratorShuffler.class";
+    public static final String hashMapKeySpliShufflerName = "java/util/HashMap$KeySpliterator$KeySpliteratorShuffler.class";
```
darko-marinov, line 76: Expand `Spli` to full name.

nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/Instrumenter.java
```diff
@@ -203,6 +206,13 @@ public byte[] apply() {
                         }
                     });
         }
+        // add SpliteratorShuffler.class
```
darko-marinov, line 209: Should this be done only in some cases, like the `if-then-else` above?

nondex-core/src/main/java/java/util/HashMap.java
```diff
@@ -1559,6 +1567,10 @@ public void forEachRemaining(Consumer<? super K> action) {
         }
 
         public boolean tryAdvance(Consumer<? super K> action) {
+            return shuffler.tryAdvance(action);
```
darko-marinov, line 1570: Good finding that you need to change this method as well.

darko-marinov (changes requested):
Thanks for adding `tryAdvance`. I've added some minor comments. I hope Kaiyao can review more, but you should also collect some experimental results while we review.

everbrightw:
Added support for instrumenting both `tryAdvance` and `forEachRemaining` methods in the `Spliterator`. Sequential stream uses both tryAdvance and forEachRemaining in different use cases. Previously, instrumenting only forEachRemaining led to undesired behavior, as the original implementation relies on a shared currentIndex to track the pipeline’s progress.

e.g.
**forEachRemaining**: 
```
keySet.stream().forEach(); // Processes elements in batches
```

**tryAdvance**:
```
keySet.stream().anyMatch(x -> x == 5); // Processes elements one at a time
```

darko-marinov:
Please rerun with the updated version as it'll help (1) automatically remove false alarms due to `tryActive` and (2) identify which tests are really `ID` (vs. `NOD` but just happened to fail by chance). Let me know if you need more VMs to run your experiments.

everbrightw:
Access to more VMs would be really helpful! Thanks!

nondex-instrumentation/src/main/java/edu/illinois/nondex/instr/Instrumenter.java
everbrightw: I can confirm for this `key` case, we do not have to do this.

We probably will need this for the `EntrySpliterator`, but I will look into this to confirm further.

